PR #104, opened 2024-04-01
chore: Add analytics setup

dbolson:
Send async analytics events to our own system to proxy instead of using vendor-specific code.


sunnyguduru (approved):

.github/actions/publish/action.yml
```diff
@@ -15,7 +15,7 @@ inputs:
     description: 'Tag to upload artifacts to.'
     required: true
 outputs:
-  hashes: 
+  hashes:
```
dbolson, line 18: These are removing trailing newlines since my editor does that automatically.

cmd/environments/get.go
```diff
@@ -61,6 +68,15 @@ func runGet(client environments.Client) func(*cobra.Command, []string) error {
 			return err
 		}
 
+		analyticsTracker.SendEvent(
```
dbolson, line 71: This is an example event. We're going to come up with a list of events to track and will implement those in another PR.

internal/analytics/client.go
```diff
@@ -0,0 +1,89 @@
...
+		Properties map[string]interface{} `json:"properties"`
+	}{
+		Event:      eventName,
+		Properties: properties,
+	}
+
+	c.wg.Add(1)
+	body, err := json.Marshal(input)
+	if err != nil { //nolint:staticcheck
+		// TODO: log error
```
dbolson, line 44: We'll probably want to add debugging at some point with a flag to enable it.

main.go
```diff
@@ -1,12 +1,23 @@
...
 )
 
 func main() {
-	cmd.Execute(version)
+	httpClient := &http.Client{
+		Timeout: time.Second * 3,
+	}
+	analyticsClient := &analytics.Client{HTTPClient: httpClient}
+	cmd.Execute(analyticsClient, version)
+	analyticsClient.Wait()
```
dbolson, line 22: This is created here so we can call `Wait()` to block the analytics request after everything else has finished.

k3llymariee (approved):

.github/actions/publish/action.yml
```diff
@@ -39,9 +39,8 @@ runs:
         args: release ${{ inputs.dry-run == 'true' && '--skip=publish' || '' }}
       env:
         GITHUB_TOKEN: ${{ inputs.token }}
-        HOMEBREW_DEPLOY_KEY: ${{ inputs.homebrew-gh-secret }}
```
k3llymariee, line 42: did we mean to remove the `HOMEBREW_DEPLOY_KEY`?

dbolson: Nope. Probably a weird merge.

sunnyguduru: Oh thanks for restoring that

PR #111, opened 2024-04-03
feat: add toggle flag step

k3llymariee:
Adds the remaining functionality for the toggle flag steps. Error handling will follow in a separate PR.

### DIFF
--- a/internal/quickstart/messages.go
+++ b/internal/quickstart/messages.go
@@ -24,6 +24,25 @@ func sendErr(err error) tea.Cmd {
 	}
 }
 
+type toggledFlagMsg struct{}
+
+func sendToggleFlagMsg(client flags.Client, accessToken, baseUri, flagKey string, enabled bool) tea.Cmd {
+	return func() tea.Msg {
+		_, err := client.Update(
+			context.Background(),
+			accessToken,
+			baseUri,
+			flagKey,
+			"default",
+			flags.BuildToggleFlagPatch("test", enabled),
+		)
+		if err != nil {
+			return sendErr(err)
+		}
+		return toggledFlagMsg{}
+	}
+}
+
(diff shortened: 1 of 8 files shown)


internal/quickstart/messages.go
```diff
@@ -24,6 +24,25 @@ func sendErr(err error) tea.Cmd {
 	}
 }
 
+type toggledFlagMsg struct{}
+
+func sendToggleFlagMsg(client flags.Client, accessToken, baseUri, flagKey string, enabled bool) tea.Cmd {
```
dbolson, line 29: Does it make more sense to name this after the message it sends, so in this case it would be `sendToggledFlagMsg` (past tense)? I could see it implicitly sending a `toggleFlagMsg` and returning the results, either a `toggledFlagMsg` or `errMsg`, but I could also see that as confusing because there isn't a specific `toggleFlagMsg` type.

internal/quickstart/toggle_flag.go
```diff
@@ -1,18 +1,30 @@
...
 	"github.com/charmbracelet/lipgloss"
+	"ldcli/internal/flags"
 )
 
 type toggleFlagModel struct {
-	flagKey string
+	accessToken string
+	baseUri     string
+	client      flags.Client
+	enabled     bool // whether the flag has ever been toggled on
```
dbolson, line 15: Could you rename this something like `flagWasToggled` since I think that `enabled` means the flag is currently toggled on and not that it was changed at all.

dbolson (approved):

internal/quickstart/messages.go
k3llymariee: ya I did actually get a bit tripped on this looking for the the actual msg type it specified when I picked it back up again, but it is also Doing The Thing here

PR #114, opened 2024-04-03
fix: remove create flag placeholder text

k3llymariee:
removes the placeholder text in the text input model and instead moves it up to the title. 
image

### DIFF
--- a/internal/quickstart/create_flag.go
+++ b/internal/quickstart/create_flag.go
@@ -10,7 +10,7 @@ import (
 	"github.com/charmbracelet/lipgloss"
 )
 
-const defaultFlagName = "my new flag"
+const defaultFlagName = "My New Flag"
 
 type createFlagModel struct {
 	accessToken string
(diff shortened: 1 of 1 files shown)


internal/quickstart/create_flag.go
```diff
@@ -69,7 +68,8 @@ func (m createFlagModel) View() string {
 		MarginLeft(2)
 
 	return fmt.Sprintf(
-		"Name your first feature flag (enter for default value):\n\n%s",
+		"Name your first feature flag (enter for default value `%s`):\n\n%s",
```
dbolson, line 71: Could you use double quotes (maybe %q would work better) instead of backticks?

dbolson (approved):
Just confirming this works the same as with the placeholder set, so if you press enter it uses the default value?

k3llymariee:
> Just confirming this works the same as with the placeholder set, so if you press enter it uses the default value?

yup we have this in there already:
```
if input == "" {
     input = defaultFlagName
}
```

PR #117, opened 2024-04-03
fix: add instructions to continue from show sdk step

k3llymariee:
image


dbolson (approved):
Should we use "press" instead of "hit" like in the toggle flag view?

PR #124, opened 2024-04-03
chore: Set up iota to describe steps instead of using ints

dbolson:
Using iota should be easier to read than magic numbers for looking at which step we're on in the setup command.

### DIFF
--- a/internal/quickstart/container.go
+++ b/internal/quickstart/container.go
@@ -16,6 +16,14 @@ const (
 	defaultEnvKey  = "test"
 )
 
+const (
+	_ = iota
+	stepCreateFlag
+	stepChooseSDK
+	stepShowSDKInstructions
+	stepToggleFlag
+)
+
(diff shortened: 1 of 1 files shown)


internal/quickstart/container.go
```diff
@@ -57,7 +65,7 @@ func (m ContainerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			cmd = tea.Quit
 		case key.Matches(msg, keys.Back):
 			// if showing SDK instructions, let the user go back to choose a different SDK
-			if m.currentStep == 2 {
+			if m.currentStep == stepShowSDKInstructions {
```
dbolson, line 68: 2 is wrong. It should be 3 which is the show SDK instructions page.

k3llymariee: yay the iotas were nice for this

k3llymariee (approved):

PR #127, opened 2024-04-04
fix: Hide error messages after successful message

dbolson:
We show an error message when appropriate, but when they are no longer relevant we should hide it.


internal/quickstart/container.go
```diff
@@ -90,6 +90,7 @@ func (m ContainerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.currentModel = NewChooseSDKModel(0)
 		m.flagKey = msg.flagKey // TODO: figure out if we maintain state here or pass in another message
 		m.currentStep += 1
+		m.err = nil
```
dbolson, line 93: These two messages are sent after a possible error has been fixed, so we should hide the error if it's still set.

k3llymariee (approved):

PR #131, opened 2024-04-04
chore: retrieve docker user and token from aws and login to docker

sunnyguduru:
We want to create and publish docker images. To do so we need to retrieve the docker use and token so that we can upload our image.

### DIFF
--- a/.github/actions/publish/action.yml
+++ b/.github/actions/publish/action.yml
@@ -21,9 +21,6 @@ runs:
       uses: docker/setup-buildx-action@v3
       with:
         platforms: linux/amd64,linux/arm64/v8,linux/arm/v7,linux/386
-    - name: Git clean
-      shell: bash
-      run: git clean -f
     - name: Login to Docker
       shell: bash
       run: |
--- a/.github/actions/publish/action.yml
+++ b/.github/actions/publish/action.yml
@@ -15,14 +15,19 @@ inputs:
 runs:
   using: composite
   steps:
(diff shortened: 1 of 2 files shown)


.github/actions/publish/action.yml
```diff
@@ -15,14 +15,19 @@ inputs:
...
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+    - name: Setup Docker Buildx
+      uses: docker/setup-buildx-action@v3
       with:
-        fetch-depth: 0
-    - name: Set up Go
-      uses: actions/setup-go@v5
-      with:
-        go-version: stable
```
sunnyguduru, line 25: Looking at ld-relay. I'm fairly confident we don't need either of these steps

cwaldren-ld: If the calling workflow has a checkout step, you should be able to omit it here. It might even be useless in the Relay flow. 

The setup Go is because we run unit tests + build Relay before publishing

.github/actions/publish/action.yml
```diff
@@ -15,14 +15,19 @@ inputs:
...
+        platforms: linux/amd64,linux/arm64/v8,linux/arm/v7,linux/386
+    - name: Git clean
```
cwaldren-ld, line 24: You can probably try omitting the `git clean` step too

cwaldren-ld (approved):

PR #132, opened 2024-04-04
chore: add arm64 docker image

sunnyguduru:
Story details: https://app.shortcut.com/launchdarkly/story/239334

Added arm64 image as well. Changed some formatting

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -25,6 +25,24 @@ dockers:
     build_flag_templates:
     - "--pull"
     - "--platform=linux/arm64/v8"
+docker_manifests:
+  - name_template: "launchdarkly/ldcli:{{ .Version}}"
+    skip_push: false
+    image_templates:
+     - "launchdarkly/ldcli:{{ .Version }}-amd64"
+     - "launchdarkly/ldcli:{{ .Version }}-arm64v8"
+
+  - name_template: "launchdarkly/ldcli:v{{ .Major }}"
+    skip_push: false
+    image_templates:
+      - "launchdarkly/ldcli:v{{ .Major }}-amd64"
+      - "launchdarkly/ldcli:v{{ .Major }}-arm64v8"
+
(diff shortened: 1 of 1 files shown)


cwaldren-ld:
If you are publishing multiple architectures, it should be as a multi-arch image. That way, they can simply pull ldcli and it will work on their platform. 

This is kinda message in goreleaser, but here's how it's done in Relay: (docker manifest) https://github.com/launchdarkly/ld-relay/blob/v8/.goreleaser.yml#L122

sunnyguduru:
> If you are publishing multiple architectures, it should be as a multi-arch image. That way, they can simply pull ldcli and it will work on their platform.
> 
> This is kinda message in goreleaser, but here's how it's done in Relay: (docker manifest) https://github.com/launchdarkly/ld-relay/blob/v8/.goreleaser.yml#L122

I see. I was wondering what that was

.goreleaser.yaml
```diff
@@ -25,7 +56,7 @@ brews:
       name: homebrew-tap
       token: "{{ .Env.GITHUB_TOKEN }}"
 
-    folder: Formula
+    directory: Formula
```
sunnyguduru, line 59: `folder` was deprecated in favor of `directory`

cwaldren-ld (approved):
Looks good. I would recommend, for times like this, that you add a ["manual publish"](https://github.com/launchdarkly/ld-relay/blob/v8/.github/workflows/manual-publish.yml) github workflow.

Essentially, it would allow you to run the publish step independently of Release Please. That's super useful if an automated release fails to release to Brew/Docker/whatever for any reason. 

Then, you can simply run the manual workflow, point it to the release tag, and try again. 

In this scenario, it would be useful because you could point it at this branch and test it out before merging. (Although since it's docker, you'd need to temporarily modify all the image names to have a "-beta.1" suffix or something similar, since they aren't "official" yet.)

PR #133, opened 2024-04-04
fix: allow users to use existing flags during setup

k3llymariee:
No longer return an error when a 409 error code is returned and instead proceed with the existing key for the remainder of the setup.

### DIFF
--- a/internal/quickstart/messages.go
+++ b/internal/quickstart/messages.go
@@ -5,11 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"ldcli/internal/environments"
-	"ldcli/internal/flags"
 	"net/http"
 
 	tea "github.com/charmbracelet/bubbletea"
+
+	"ldcli/internal/environments"
+	"ldcli/internal/errors"
+	"ldcli/internal/flags"
 )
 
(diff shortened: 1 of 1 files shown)


internal/quickstart/messages.go
```diff
@@ -59,7 +61,14 @@ func sendCreateFlagMsg(client flags.Client, accessToken, baseUri, flagName, flag
...
 		if err != nil {
-			return errMsg{err: err}
+			var e struct {
+				Code    string `json:"code"`
+				Message string `json:"message"`
+			}
+			_ = json.Unmarshal([]byte(err.Error()), &e)
+			if e.Code != "conflict" {
+				return errMsg{err: errors.NewError(fmt.Sprintf("Error creating flag: %s. Press \"ctrl + c\" to quit.", e.Message))}
+			}
```
k3llymariee, line 71: maybe we clean this up when we do this for other steps also? https://app.shortcut.com/launchdarkly/story/239181/errors-during-setup-command-should-be-human-readable?vc_group_by=day&ct_workflow=all&cf_workflow=500010659

dbolson: Yeah, I think we could get something more generic when we do that one.

dbolson (approved):

PR #134, opened 2024-04-04
feat: show success message after creating a flag

k3llymariee:
Adds an interstitial step to show the user that either a new flag was created, or an existing one was used, during the first step of the setup command. 

new-flag

existing-flag


internal/quickstart/create_flag.go
```diff
@@ -13,10 +13,13 @@ import (
...
-	accessToken string
-	baseUri     string
-	client      flags.Client
-	textInput   textinput.Model
+	accessToken      string
+	baseUri          string
+	client           flags.Client
+	existingFlagUsed bool
+	flagKey          string
+	success          bool
```
k3llymariee, line 21: these are the new fields - also storing the flag key now to pass into the view

internal/quickstart/create_flag.go
```diff
@@ -58,6 +61,14 @@ func (m createFlagModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		default:
 			m.textInput, cmd = m.textInput.Update(msg)
 		}
+	case createdFlagMsg:
+		if !m.success {
```
k3llymariee, line 65: i don't love this name so open to suggestions

internal/quickstart/create_flag.go
```diff
@@ -67,6 +78,14 @@ func (m createFlagModel) View() string {
 	style := lipgloss.NewStyle().
 		MarginLeft(2)
 
+	if m.success {
+		successMessage := fmt.Sprintf("Flag %q created successfully!", m.flagKey)
+		if m.existingFlagUsed {
+			successMessage = fmt.Sprintf("Using existing existing flag %q for setup.", m.flagKey)
```
dbolson, line 84: Stutter with "existing."

internal/quickstart/create_flag.go
```diff
@@ -67,6 +78,14 @@ func (m createFlagModel) View() string {
 	style := lipgloss.NewStyle().
 		MarginLeft(2)
 
+	if m.success {
+		successMessage := fmt.Sprintf("Flag %q created successfully!", m.flagKey)
```
dbolson, line 82: Should we use the flag name instead of the key since that's what the user typed? Otherwise they could be surprised if they type "my flag" and we show "my-flag." If we go that route it may be simpler to create a flag struct for the key and name instead of passing both around.

internal/quickstart/create_flag.go
dbolson: Does "success" mean that the user created a new flag or tried to create an existing flag so that we would (over)write the flag key with the one from the message? Maybe instead of "success" it could be something like "shouldUpdateFlag"?

internal/quickstart/create_flag.go
```diff
@@ -67,6 +85,14 @@ func (m createFlagModel) View() string {
 	style := lipgloss.NewStyle().
 		MarginLeft(2)
 
+	if m.showSuccessView {
+		successMessage := fmt.Sprintf("Flag %q created successfully!", m.flag.name)
+		if m.existingFlagUsed {
+			successMessage = fmt.Sprintf("Using existing flag %q for setup.", m.flag.name)
+		}
+		return successMessage + " Press enter to continue."
+	}
```
k3llymariee, line 94: @dbolson updated the name here - is this doing too much/do I need another message type here to pass around?

dbolson: Yeah, maybe we should have a message for when we create a new flag and one where the flag already exists.

k3llymariee: ah I meant for this comment to go on the Update step where I was overloading one message type, but think I've managed to fix

internal/quickstart/create_flag.go
```diff
@@ -14,14 +14,22 @@ import (
 
 const defaultFlagName = "My New Flag"
 
+type flagDetail struct {
+	key  string
+	name string
+}
+
```
k3llymariee, line 21: added new flagDetail type

internal/quickstart/container.go
```diff
@@ -69,7 +69,9 @@ func (m ContainerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		case key.Matches(msg, pressableKeys.Back):
 			switch m.currentStep {
 			case stepCreateFlag:
-				// can't go back
+				// can only go back if a flag has been created but not confirmed,
+				// so let the model handle the Update
+				m.currentModel, cmd = m.currentModel.Update(msg)
```
k3llymariee, line 74: this is new also, to allow going back after a flag has been created (but not before)

internal/quickstart/create_flag.go
```diff
@@ -63,9 +75,18 @@ func (m createFlagModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+				m.showSuccessView = false
+			}
 		default:
 			m.textInput, cmd = m.textInput.Update(msg)
 		}
+	case createdFlagMsg:
+		m.showSuccessView = true
+		m.existingFlagUsed = msg.existingFlagUsed
+		m.flag = msg.flag
+		return m, cmd
```
dbolson, line 89: I'm not sure this needs to return early.

internal/quickstart/create_flag.go
```diff
@@ -14,14 +14,22 @@ import (
 
 const defaultFlagName = "My New Flag"
 
+type flagDetail struct {
```
dbolson, line 17: We could probably only need to call this `flag`.

dbolson (approved):

PR #135, opened 2024-04-04
fix: create flag prompt should be on the same line as text

k3llymariee:
before:
different-line

after:
same-line-input

### DIFF
--- a/internal/quickstart/create_flag.go
+++ b/internal/quickstart/create_flag.go
@@ -24,6 +24,7 @@ func NewCreateFlagModel(client flags.Client, accessToken, baseUri string) tea.Mo
 	ti.Focus()
 	ti.CharLimit = 156
 	ti.Width = 20
+	ti.Prompt = ""
 
 	return createFlagModel{
 		accessToken: accessToken,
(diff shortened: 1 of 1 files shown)


internal/quickstart/create_flag.go
```diff
@@ -24,6 +24,7 @@ func NewCreateFlagModel(client flags.Client, accessToken, baseUri string) tea.Mo
 	ti.Focus()
 	ti.CharLimit = 156
 	ti.Width = 20
+	ti.Prompt = ""
```
k3llymariee, line 27: this gets rid of the ">"

internal/quickstart/create_flag.go
```diff
@@ -68,8 +69,7 @@ func (m createFlagModel) View() string {
 		MarginLeft(2)
 
 	return fmt.Sprintf(
-		"Name your first feature flag (enter for default value %q):\n\n%s",
-		defaultFlagName,
```
k3llymariee, line 72: i removed this because the prompt felt a little long to have on one line with the input. plus with the new confirmation step they'll see the value that was used.

sunnyguduru (approved):
Nice!

dbolson (approved):

PR #137, opened 2024-04-05
chore: update access-token flag help text

k3llymariee:
image

Adding a link to the account settings page felt like it would be too long.


dbolson (approved):
Do you think saying "Your LaunchDarkly API token with write-level access" would make it more obvious? I could go either way.

k3llymariee:
> Do you think saying "Your LaunchDarkly API token with write-level access" would make it more obvious? I could go either way.

hmm I think I'm going to leave it as-is for now and maybe wait for more feedback from users?

PR #138, opened 2024-04-05
feat: show help text

dbolson:
Show help text for every step. We want to support going back for each step before merging this.

Also, we should try to fix the padding in the list of SDKs separate from this change.

Screenshot 2024-04-05 at 10 06 55 AM
Screenshot 2024-04-05 at 10 07 00 AM
Screenshot 2024-04-05 at 10 07 09 AM
Screenshot 2024-04-05 at 10 07 15 AM


internal/quickstart/create_flag.go
```diff
@@ -29,7 +33,18 @@ func NewCreateFlagModel(client flags.Client, accessToken, baseUri string) tea.Mo
 		accessToken: accessToken,
 		baseUri:     baseUri,
 		client:      client,
-		textInput:   ti,
+		help:        help.New(),
+		helpKeys: keyMap{
+			Back: key.NewBinding(
+				key.WithKeys("esc"),
+				key.WithHelp("esc", "back"),
+			),
```
k3llymariee, line 41: we can't go back on the input step here though - with the new success page we will be able to but is there a way to hide it when we should the textinput View?

internal/quickstart/toggle_flag.go
```diff
@@ -75,5 +92,5 @@ func (m toggleFlagModel) View() string {
 		Padding(0, 1).
 		MarginRight(margin)
 
-	return title + "\n\n" + toggleStyle.Render(toggle) + m.flagKey + furtherInstructions
+	return title + "\n\n" + toggleStyle.Render(toggle) + m.flagKey + furtherInstructions + "\n\n" + helpView
```
k3llymariee, line 95: the help text jumps after the the flag is toggled for the first time. I don't think there's an easy way to pad this to make sure that doesn't happen though 😞

k3llymariee (approved):

internal/quickstart/toggle_flag.go
dbolson: I was able to add some padding to keep the help text on the same line, but I'm not sure if the additional space is better than the jump. Maybe we leave the jump to see if anybody comments on it?
Screenshot 2024-04-05 at 1 56 24 PM

k3llymariee: sgtm!

internal/quickstart/create_flag.go
dbolson: I removed "back" from the help text for create flag.

PR #145, opened 2024-04-08
build: create manual publish gh workflow

sunnyguduru:
Added a manual-publish.yml file to create the manual-publish workflow

### DIFF
--- a/.github/workflows/manual-publish.yml
+++ b/.github/workflows/manual-publish.yml
@@ -36,7 +36,7 @@ jobs:
       - uses: ./.github/actions/publish
         with:
           dry-run: ${{ inputs.dry_run }}
-          token: ${{ var.ldcli_homebrew-tap_deploy_key }}
+          token: ${{ secrets.GITHUB_TOKEN }}
           tag: ${{ inputs.tag }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-publish.yml
```diff
@@ -0,0 +1,52 @@
...
+          fetch-depth: 0
+          
+      - uses: launchdarkly/gh-actions/actions/release-secrets@release-secrets-v1.0.1
+        name: 'Get Docker token'
+        with:
+          aws_assume_role: ${{ vars.AWS_ROLE_ARN }}
+          ssm_parameter_pairs: '/global/services/docker/public/username = DOCKER_USERNAME, /global/services/docker/public/token = DOCKER_TOKEN'
+          
+      - uses: ./.github/actions/publish
+        with:
```
rsoberano-ld, line 37: I think we need to add an `id: publish` here so that `${{ steps.publish.outputs.hashes}}` above works

sunnyguduru: Added!

rsoberano-ld (approved):

PR #150, opened 2024-04-09
feat: make sdk instructions scrollable

k3llymariee:
https://github.com/launchdarkly/ldcli/assets/55991524/fe35ac14-d33e-41b8-8d03-851176469058

Adds a viewport for displaying the SDK instructions, which should be scrollable both with up/down keys as well as a mouse wheel.

### DIFF
--- a/internal/quickstart/container.go
+++ b/internal/quickstart/container.go
@@ -110,6 +110,7 @@ func (m ContainerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.flagKey = msg.flagKey // TODO: figure out if we maintain state here or pass in another message
 		m.currentStep += 1
 		m.err = nil
+		cmd = sendEnableMouseCellMotionMsg()
 	case errMsg:
 		m.currentModel, cmd = m.currentModel.Update(msg)
 	case noInstructionsMsg:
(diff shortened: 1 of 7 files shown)


internal/quickstart/container.go
```diff
@@ -112,6 +112,7 @@ func (m ContainerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.flagKey = msg.flag.key
 		m.currentStep += 1
 		m.err = nil
+		cmd = sendEnableMouseCellMotionMsg()
```
k3llymariee, line 115: we can either enable this here as a message or at the start of the quickstart program, thought it was safer to enable here?

internal/quickstart/show_sdk_instructions.go
```diff
@@ -2,18 +2,21 @@ package quickstart
...
 	"github.com/charmbracelet/lipgloss"
-	"github.com/muesli/reflow/wordwrap"
-
 	"ldcli/internal/sdks"
 )
 
+const (
+	viewportWidth  = 80
+	viewportHeight = 30
+)
```
k3llymariee, line 18: arbitrary values I picked - we could adjust these based on the window size if we want?

internal/quickstart/container.go
sunnyguduru: I think it'd better here too. Since we don't use the feature elsewhere

internal/quickstart/show_sdk_instructions.go
```diff
@@ -71,12 +83,21 @@ func (m showSDKInstructionsModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 	case fetchedSDKInstructions:
 		m.instructions = sdks.ReplaceFlagKey(string(msg.instructions), m.flagKey)
 	case fetchedEnv:
 		m.sdkKey = msg.sdkKey
 		m.instructions = sdks.ReplaceSDKKey(string(m.instructions), msg.sdkKey)
+		md, err := m.renderMarkdown()
+		if err != nil {
+			return m, sendErr(err)
+		}
+		m.viewport.SetContent(md)
```
k3llymariee, line 100: rendering once here vs. every time the `View` method is called, to allow for proper updating

sunnyguduru (approved):

dbolson (approved):

PR #151, opened 2024-04-09
build: deploy formula to homebrew-tap

sunnyguduru:
Tested successfully here in a [manual publish run](https://github.com/launchdarkly/ldcli/actions/runs/8608236726)


rsoberano-ld (approved):

.github/actions/publish/action.yml
```diff
@@ -4,6 +4,9 @@ inputs:
   token:
     description: 'Token to use for publishing.'
     required: true
+  homebrew-gh-secret:
+    description: 'Token to use for publishing to homebrew-tap.'
```
rsoberano-ld, line 8: Technically an SSH private key and not a token, we should probably clarify that here

PR #152, opened 2024-04-09
feat: Use new sdk instructions instead of hello READMEs

k3llymariee:
Copies the instructions from the guided quickstart [here](https://github.com/launchdarkly/gonfalon/tree/main/static/ld/components/getStarted/sdk) as markdown files to render in the setup flow. Starting with just one SDK (python) to establish the pattern. 

image

### DIFF
--- a/internal/quickstart/messages.go
+++ b/internal/quickstart/messages.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 
 	tea "github.com/charmbracelet/bubbletea"
 
@@ -131,6 +132,17 @@ func sendFetchSDKInstructionsMsg(url string) tea.Cmd {
 		}
 
 		return fetchedSDKInstructions{instructions: body}
+	}
+}
+
+func sendReadSDKInstructionsMsg(filename string) tea.Cmd {
+	return func() tea.Msg {
+		content, err := os.ReadFile(fmt.Sprintf("internal/sdks/sdk_instructions/%s.md", filename))
+		if err != nil {
+			return errMsg{err: err}
+		}
+
+		return fetchedSDKInstructions{instructions: content}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


internal/sdks/sdk_instructions/python.md
```diff
@@ -0,0 +1,56 @@
...
+
+```python
+# Import the LaunchDarkly client.
+import ldclient
+from ldclient import Context
+from ldclient.config import Config
+
+# Create a helper function for rendering messages.
+def show_message(s):
+    print("*** {}".format(s))
```
k3llymariee, line 28: I _did_ change the string formatting from what's shown in the quickstart because our rendering was trying/failing to find the `%s` values and instead was showing `MISSING` values

before:
```python
print("*** %s" % s)
```
image

internal/quickstart/messages.go
```diff
@@ -134,6 +135,17 @@ func sendFetchSDKInstructionsMsg(url string) tea.Cmd {
...
 }
 
+func sendReadSDKInstructionsMsg(filename string) tea.Cmd {
+	return func() tea.Msg {
+		content, err := os.ReadFile(fmt.Sprintf("internal/sdks/sdk_instructions/%s.md", filename))
+		if err != nil {
+			return errMsg{err: err}
+		}
+
+		return fetchedSDKInstructions{instructions: content}
```
k3llymariee, line 145: despite the name of the function we're still sending `fetchedSDKInstructions` message type - figured we can update either the message name or the function name when we stop fetching from the READMEs but this way we can still just handle this one message in the `showSDKInstructionsModel.Update` method.

internal/sdks/sdk_instructions/python.md
```diff
@@ -0,0 +1,56 @@
+# Set up your application
+
+If you want to skip ahead, the final code is available in our [GitHub repository](https://github.com/launchdarkly/hello-python).
+
+
```
k3llymariee, line 5: I think maybe when the instructions are all in we might want to pull this intro text out of the markdown file/viewport box

sunnyguduru (approved):

dbolson (approved):

PR #155, opened 2024-04-09
fix: space above pagination dots and pagination bug

sunnyguduru:
I removed the extra space between the choose sdk list and the pagination dots.
CleanShot 2024-04-09 at 15 56 58@2x

bug fix: we were only showing 15 items total. This was due to setting pagination.PerPage manually to 5. We can't manually control this property without bugs due to how list.updatePagination() works. Setting the list height to 9 shows us 5 items, and displays all 24-25 items.

video: https://share.cleanshot.com/bs9YgdBF

### DIFF
--- a/internal/quickstart/choose_sdk.go
+++ b/internal/quickstart/choose_sdk.go
@@ -31,7 +31,7 @@ type chooseSDKModel struct {
 }
 
 func NewChooseSDKModel(selectedIndex int) tea.Model {
-	l := list.New(sdksToItems(), sdkDelegate{}, 30, 14)
+	l := list.New(sdksToItems(), sdkDelegate{}, 20, 9)
 	l.Title = "Select your SDK:\n"
 	// reset title styles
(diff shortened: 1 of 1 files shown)


internal/quickstart/choose_sdk.go
```diff
@@ -40,7 +40,6 @@ func NewChooseSDKModel(selectedIndex int) tea.Model {
 	l.SetShowPagination(true)
 	l.SetShowStatusBar(false)
 	l.SetFilteringEnabled(false) // TODO: try to get filtering working
-	l.Paginator.PerPage = 5
```
sunnyguduru, line 43: We're not allowed to manually set this unfortunately. The property gets computed automatically by list.updatePagination(). So setting it manually messes up the computation. I found a way to show exactly 5 items or less on every page though by setting the hight on line 34 to 9.

dbolson (approved):

PR #157, opened 2024-04-10
chore: add react sdk instructions

k3llymariee:
Adding this one as its own PR because it's a little more involved than just adding the `.md` file. 

react

### DIFF
--- a/internal/quickstart/messages.go
+++ b/internal/quickstart/messages.go
@@ -161,7 +161,8 @@ func sendEnableMouseCellMotionMsg() tea.Cmd {
 }
 
 type fetchedEnv struct {
-	sdkKey string
+	clientSideId string
+	sdkKey       string
 }
 
 func sendFetchEnv(accessToken string, baseUri string, key string, projKey string) tea.Cmd {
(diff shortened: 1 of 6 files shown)


internal/quickstart/messages.go
```diff
@@ -173,14 +174,15 @@ func sendFetchEnv(accessToken string, baseUri string, key string, projKey string
 		}
 
 		var resp struct {
-			SDKKey string `json:"apiKey"`
+			SDKKey       string `json:"apiKey"`
+			ClientSideId string `json:"_id"`
```
k3llymariee, line 178: client side sdks use client side ids 🙃

internal/sdks/sdks.go
```diff
@@ -23,21 +27,33 @@ func ReplaceFlagKey(instructions string, key string) string {
...
-		key,
+		sdkKey,
+		"myClientSideID",
+		clientSideId,
 	)
 
 	return r.Replace(instructions)
 }
+
+// kebabToCamel converts a kebab-case key string into a camelCase key string, used for the React sdk instructions
```
k3llymariee, line 52: from [the docs](https://docs.launchdarkly.com/sdk/client-side/react/react-web?site=launchDarkly#flag-keys-in-the-react-web-sdk):

>However, in JavaScript, accessing keys containing . and - requires the bracket operator. Instead of requiring this, the React Web SDK automatically changes all flag keys to camel case by default. A flag with key dev-flag-test is accessible as flags.devFlagTest.

so we should do that here also

internal/sdks/sdks.go
```diff
@@ -23,21 +27,33 @@ func ReplaceFlagKey(instructions string, key string) string {
 	return r.Replace(instructions)
 }
 
-// ReplaceSDKKey changes the placeholder flag key in the SDK instructions to the flag key from
-// the user.
-func ReplaceSDKKey(instructions string, key string) string {
+// ReplaceSDKKeys changes the placeholder SDK key/client side ID in the SDK instructions to the key from
+// the default test environment for the user's account.
+func ReplaceSDKKeys(instructions string, sdkKey, clientSideId string) string {
```
k3llymariee, line 32: updated this function name and signature to update both the SDK key and client side ID

sunnyguduru (approved):

PR #160, opened 2024-04-10
Removing mistaken SDK references

bwoskow-ld:
It looks like the provenance content was copy/pasted from our SDK repos, but this isn't an SDK.


rsoberano-ld (approved):
Thanks for catching this Ben!

sunnyguduru (approved):

sunnyguduru:
Renamed to build to that it doesn't show up in the changelog

PR #161, opened 2024-04-10
fix: remove mouse wheel support from show sdk & show scroll options in help view

k3llymariee:
Removes mouse wheel support from the viewport, because with this enabled we're unable to highlight the text within to copy/paste (which is kinda the whole point of it). 

image

### DIFF
--- a/internal/quickstart/show_sdk_instructions.go
+++ b/internal/quickstart/show_sdk_instructions.go
@@ -94,8 +94,6 @@ func (m showSDKInstructionsModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		default:
 			m.viewport, cmd = m.viewport.Update(msg)
 		}
-	case tea.MouseMsg:
-		m.viewport, cmd = m.viewport.Update(msg)
 	case fetchedSDKInstructions:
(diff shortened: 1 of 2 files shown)


internal/quickstart/show_sdk_instructions.go
```diff
@@ -51,16 +51,21 @@ func NewShowSDKInstructionsModel(
 		BorderForeground(lipgloss.Color("62")).
 		PaddingRight(2)
 
+	h := help.New()
+	h.ShowAll = true
```
sunnyguduru, line 55: Did we define help ourselves somewhere?

sunnyguduru (approved):

dbolson (approved):
The full version of the help with fewer elements looks kind of weird. It could be good enough without redoing the entire help, but do you think defaulting to short help and showing "more" would hide the fact that you can scroll too much?

k3llymariee:
@dbolson yeah I worry about hiding those keys/making it less obvious that it's scrollable. I don't hate it as-is 😅 but maybe we can wait for more feedback

internal/quickstart/show_sdk_instructions.go
k3llymariee: this is just the bubbletea Help model

PR #162, opened 2024-04-10
fix: remove side borders from show SDK viewport

k3llymariee:
I noticed that when copy/pasting the sdk instructions with a border for the viewport, we end up with something like this on the clipboard: 
```
│  3. Create a file called  test.py  and add the following code:                 │
│                                                                                │
│    # Import the LaunchDarkly client.                                           │
│    import ldclient                                                             │
│    from ldclient import Context                                                │
│    from ldclient.config import Config                                          │
│                                                                                │
│    # Create a helper function for rendering messages.                          │
│    def show_message(s):                                                        │
│        print("*** {}".format(s))                                               │
│        print()                                                                 │
```

My proposed solution is to remove the side borders - with the additional help keys at the bottom it should hopefully be clear that the text is scrollable, but open to opinions. 

borders

An alternative would be to use a `HiddenBackground` with a `BorderBackground` color, which would look something like this: 
image

### DIFF
--- a/internal/quickstart/container.go
+++ b/internal/quickstart/container.go
@@ -114,7 +114,6 @@ func (m ContainerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.flagKey = msg.flag.key
 		m.currentStep += 1
 		m.err = nil
-		cmd = sendEnableMouseCellMotionMsg()
 	case errMsg:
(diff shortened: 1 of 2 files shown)


sunnyguduru (approved):

dbolson (approved):
Works for me

PR #165, opened 2024-04-10
feat: add extra instructions on first step

k3llymariee:
Per feedback, make it more clear that this command will actually be doing. 

image

### DIFF
--- a/internal/quickstart/container.go
+++ b/internal/quickstart/container.go
@@ -37,6 +37,7 @@ type ContainerModel struct {
 	err                error
 	flagKey            string
 	flagsClient        flags.Client
+	gettingStarted     bool
 	quitting           bool
 	sdk                sdkDetail
 	totalSteps         int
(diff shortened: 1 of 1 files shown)


internal/quickstart/container.go
```diff
@@ -162,5 +165,9 @@ func (m ContainerModel) View() string {
 		return ""
 	}
 
+	if m.gettingStarted {
+		out = "Within this guided setup flow, you'll be creating a new feature flag and,\nusing the SDK of your choice, building a small sample application to see a\nfeature flag toggle on and off in real time.\n\nLet's get started!\n" + out
+	}
+
```
k3llymariee, line 171: added to the container view because I wanted it above the `Step 1 of 4` bit, but let me know if we'd rather have this in the `createFlagModel` view, because that text will jump around a bit in the next step

jump-scare

dbolson (approved):

PR #168, opened 2024-04-11
chore: don't refetch envs when going back from show sdk step

k3llymariee:
no need to refetch the env when re-choosing a new SDK when we already have the keys! this makes the loading of sdk steps a lot quicker if you're checking out a bunch of different sdks and going back and forth between steps

### DIFF
--- a/internal/quickstart/show_sdk_instructions.go
+++ b/internal/quickstart/show_sdk_instructions.go
@@ -2,7 +2,6 @@ package quickstart
 
 import (
 	"fmt"
-
 	"github.com/charmbracelet/bubbles/help"
 	"github.com/charmbracelet/bubbles/key"
 	"github.com/charmbracelet/bubbles/spinner"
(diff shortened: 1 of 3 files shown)


internal/quickstart/show_sdk_instructions.go
```diff
@@ -112,10 +121,16 @@ func (m showSDKInstructionsModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			m.viewport, cmd = m.viewport.Update(msg)
 		}
 	case fetchedSDKInstructionsMsg:
-		m.instructions = sdks.ReplaceFlagKey(string(msg.instructions), m.flagKey)
```
k3llymariee, line 115: refactored these to `renderMarkdown`

sunnyguduru (approved):

internal/quickstart/show_sdk_instructions.go
```diff
@@ -20,18 +19,24 @@ const (
 	viewportHeight = 30
 )
 
+type envKeys struct {
```
dbolson, line 22: I keep getting tripped up by this name since I expect it to be a slice of the key value of multiple environments, not the various keys that are defined on an environment. What about calling it `environment` like we do with the `flag` type where it's a subset of the full data for an environment that we could add to later?

dbolson (approved):

internal/quickstart/show_sdk_instructions.go
k3llymariee: fair enough! updated

PR #170, opened 2024-04-11
chore: show human readable errors for toggle flag and fetch environment

sunnyguduru:
CleanShot 2024-04-16 at 15 47 52


internal/quickstart/messages.go
```diff
@@ -40,7 +40,12 @@ func toggleFlag(client flags.Client, accessToken, baseUri, flagKey string, enabl
 			flags.BuildToggleFlagPatch(defaultEnvKey, enabled),
 		)
 		if err != nil {
-			return errMsg{err: err}
+			var e struct {
```
dbolson, line 43: Can you make this show a nicer error message on conflict like the `createFlag` command? And then could you refactor the duplication to a single place?

sunnyguduru: Oh sure. What would the conflict on a flag toggle look like? Would it be if the flag is already targeted on and we're target it on with our api call?

dbolson: Try toggling it a bunch to see the error show up.

internal/quickstart/toggle_flag.go
```diff
@@ -82,6 +85,9 @@ func (m toggleFlagModel) View() string {
 
 	if m.flagWasEnabled {
 		furtherInstructions = fmt.Sprintf("\n\nCheck your %s to see the change!", logTypeMap[m.sdkKind])
+		if m.hasToggleConflict {
+			furtherInstructions = fmt.Sprintf("\n\nThe flag is already toggled %v.", toggle)
+		}
```
sunnyguduru, line 90: Showing this message when flag is already toggled to either "on" or "off" on the server

internal/quickstart/messages.go
```diff
@@ -203,3 +208,18 @@ func selectedSDK(index int) tea.Cmd {
...
+		Message string `json:"message"`
+	}
+	_ = json.Unmarshal([]byte(err.Error()), &e)
+	conflict := e.Code == "conflict"
+	if !conflict {
+		errorMsg = errMsg{err: errors.NewError(fmt.Sprintf("%s: %s. Press \"ctrl + c\" to quit.", message, e.Message))}
+	}
+
+	return conflict, errorMsg
+}
```
sunnyguduru, line 225: Helper function so we have less repetitive error handling code.

internal/quickstart/messages.go
```diff
@@ -40,10 +43,14 @@ func toggleFlag(client flags.Client, accessToken, baseUri, flagKey string, enabl
 			flags.BuildToggleFlagPatch(defaultEnvKey, enabled),
 		)
 		if err != nil {
-			return errMsg{err: err}
+			var errorMsg errMsg
+			hasToggleConflict, errorMsg = handleError(err, "Error toggling flag")
+			if !hasToggleConflict {
+				return errorMsg
+			}
```
sunnyguduru, line 50: I don't like that I'm checking the conflict variable here again. I don't have an alternative solution in mind for this though

sunnyguduru:
gif added to description

dbolson (commented):
I think something [like this setup](https://github.com/launchdarkly/ldcli/pull/179) will be simpler so we don't need to pass a boolean flag in the message.

internal/quickstart/messages.go
```diff
@@ -117,6 +136,7 @@ func createFlag(client flags.Client, accessToken, baseUri, flagName, flagKey, pr
 					),
 				}
 			}
+			existingFlag = true
```
sunnyguduru, line 139: I realized that with the recent changes #179 we were now forgetting to set `existingFlag` when there is a createFlag conflict

internal/quickstart/toggle_flag.go
```diff
@@ -94,6 +97,9 @@ func (m toggleFlagModel) View() string {
 
 	if m.flagWasEnabled {
 		furtherInstructions = fmt.Sprintf("\n\nCheck your %s to see the change!", logTypeMap[m.sdkKind])
+		if m.alreadyEnabled {
+			furtherInstructions = "\n\nFlag was toggled too quickly."
```
sunnyguduru, line 101: Removed toggle state. Also followed the example from line 70 `Error toggling flag: you have toggled too quickly.`

internal/quickstart/messages.go
```diff
@@ -25,10 +25,14 @@ func sendErrMsg(err error) tea.Cmd {
 	}
 }
 
-type toggledFlagMsg struct{}
+type toggledFlagMsg struct {
+	alreadyEnabled bool
```
dbolson, line 29: I'm not sure this is more clear than passing an error and letting the model handle it. The `createFlag` command decides a particular error response isn't an actual error while the `toggleFlag` command always returns an error.

If the `toggleFlag` command returns either success (`toggledFlagMsg`) or failure (`errMsg`), then the model can decide how to show the error message instead of whether it's an error or not like the `createFlag` command.

sunnyguduru: That's fair. I'll update to follow that format

internal/quickstart/messages.go
```diff
@@ -189,7 +200,21 @@ func fetchEnv(
 		}
 		err = json.Unmarshal(response, &resp)
 		if err != nil {
-			return errMsg{err: err}
+			msgRequestErr, err := newMsgRequestError(err.Error())
+			if err != nil {
+				return errMsg{err: err}
+			}
+
+			if !msgRequestErr.IsConflict() {
```
dbolson, line 208: I don't think we can get a conflict error when fetching the environment, so we're probably fine to just return the error every time.

internal/quickstart/messages.go
```diff
@@ -38,16 +38,29 @@ func toggleFlag(client flags.Client, accessToken, baseUri, flagKey string, enabl
 			flags.BuildToggleFlagPatch(defaultEnvKey, enabled),
 		)
 		if err != nil {
-			return errMsg{err: err}
+			msgRequestErr, err := newMsgRequestError(err.Error())
+			if err != nil {
+				return errMsg{err: err}
+			}
+
+			if !msgRequestErr.IsConflict() {
```
dbolson, line 46: I don't think we need this since we're checking for this in the model.

dbolson (commented):
I get this message when I try to use "new" for the flag name:
> invalid character 'E' looking for beginning of value

I also see "Error toggling flag: you have toggled too quickly" occasionally in addition to "Flag was toggled too quickly." Can you manually check these fixes?

sunnyguduru:
> I get this message when I try to use "new" for the flag name:
> 
> > invalid character 'E' looking for beginning of value
> 
> I also see "Error toggling flag: you have toggled too quickly" occasionally in addition to "Flag was toggled too quickly." Can you manually check these fixes?

Yeah, I can look into that. Weird I didn't see the E error with `Sunny test flag`.

sunnyguduru:
> I get this message when I try to use "new" for the flag name:
> 
> > invalid character 'E' looking for beginning of value
> 
> I also see "Error toggling flag: you have toggled too quickly" occasionally in addition to "Flag was toggled too quickly." Can you manually check these fixes?

The first issue is related to footerView. Seems that we were originally displaying errors there. I think it seems valuable to continue to show errors through footer view. So I want to remove showing errors through the "further instructions" string

sunnyguduru:
> > I get this message when I try to use "new" for the flag name:
> > > invalid character 'E' looking for beginning of value
> > 
> > 
> > I also see "Error toggling flag: you have toggled too quickly" occasionally in addition to "Flag was toggled too quickly." Can you manually check these fixes?
> 
> The first issue is related to footerView. Seems that we were originally displaying errors there. I think it seems valuable to continue to show errors through footer view. So I want to remove showing errors through the "further instructions" string

Although the spacing gets more confusing that way

sunnyguduru:
Fixed how errors was not being displayed correctly on the create flag page:
CleanShot 2024-04-15 at 16 24 00@2x

internal/quickstart/messages.go
```diff
@@ -122,7 +119,7 @@ func createFlag(client flags.Client, accessToken, baseUri, flagName, flagKey, pr
 		return createdFlagMsg{flag: flag{
 			key:  flagKey,
 			name: flagName,
-		}, existingFlagUsed: existingFlag}
```
sunnyguduru, line 125: exisitingFlagUsed wasn't getting used anymore so I removed references to it.

internal/quickstart/toggle_flag.go
```diff
@@ -92,7 +92,7 @@ func (m toggleFlagModel) View() string {
 		toggle = "ON"
 	}
 
-	if m.flagWasEnabled {
+	if m.flagWasEnabled && m.err == nil {
 		furtherInstructions = fmt.Sprintf("\n\nCheck your %s to see the change!", logTypeMap[m.sdkKind])
 	}
```
sunnyguduru, line 97: Prevent showing the success message when there is an error.

sunnyguduru (commented):
here's a gif of me testing the changes:
- Errors show up correctly for create flag (as well as conflict)
- Errors show up correctly for toggle flag
CleanShot 2024-04-16 at 15 47 52

PR #185, opened 2024-04-17
fix: Change config file location

dbolson:
Running `ldcli config --list`, `--set foo bar`, or `--unset foo` without a config file should not error now.

If we keep it in the home directory instead of in a nested one, we don't need to worry about creating the directory if it does not exist.

**Requirements**

**Related issues**

**Additional context**

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ LaunchDarkly CLI commands
 - `setup` guides you through creating your first flag, connecting an SDK, and evaluating your flag in your Test environment
 
 ### Resource Commands
+
 Resource commands mirror the LaunchDarkly API and make requests for a given resource. To see a full list of resources supported by the CLI, enter `ldcli --help` into your terminal, and to see the commands available for a given resource
 
 To see the commands available for a given resource:
@@ -52,6 +53,7 @@ _(coming soon!)_
 
 [//]: # (TODO: add info about how to opt out)
 
+Running `ldcli config --set access-token {your access token}` will create a configuration file located at `$HOME/.ldcli-config.yml` with the access token. Further commands will read from this file so you do not need to specify the access token each time.
 
 ## Contributing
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -52,14 +53,15 @@ _(coming soon!)_
 
 [//]: # (TODO: add info about how to opt out)
 
+Running `ldcli config --set access-token {your access token}` will create a configuration file located at `$HOME/.ldcli-config.yml` with the access token. Further commands will read from this file so you do not need to specify the access token each time.
```
dbolson, line 56: Added this in case people want to look at the file.

cmd/config/config.go
```diff
@@ -53,6 +53,10 @@ func run() func(*cobra.Command, []string) error {
 				return err
 			}
 
+			if string(configJSON) == "{}" {
```
dbolson, line 56: Empty yaml is `{}`, but it looks weird to show that.

cmd/config/config.go
```diff
@@ -146,18 +150,14 @@ func getRawConfig() (map[string]interface{}, *viper.Viper, error) {
...
-		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
-			newViper := viper.New()
-			newViper.SetConfigFile(config.GetConfigFile())
-			err = newViper.WriteConfigAs(config.GetConfigFile())
-			if err != nil {
-				return nil, err
-			}
-
-			return newViper, nil
+		newViper := viper.New()
```
dbolson, line 153: It simplifies things to assume any reasonable error is because the file doesn't exist, so we'll try to write a new one regardless of the type of error returned.

cmd/root.go
```diff
@@ -143,18 +140,7 @@ func Execute(analyticsTracker analytics.Tracker, version string) {
...
-		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
-			// ignore if file not found
-		} else {
-			return err
-		}
-	}
-
-	return nil
+func setFlagsFromConfig() {
+	viper.SetConfigFile(config.GetConfigFile())
```
dbolson, line 144: `SetConfigFile` is the same as the other lines.

internal/config/config.go
```diff
@@ -35,20 +33,15 @@ func NewConfig(rawConfig map[string]interface{}) ConfigFile {
...
 
 func GetConfigPath() string {
-	configPath := os.Getenv("XDG_CONFIG_HOME")
-	if configPath == "" {
-		home, err := homedir.Dir()
-		if err != nil {
-			return ""
-		}
-		configPath = filepath.Join(home, ".config")
+	home, err := homedir.Dir()
```
dbolson, line 36: The main issue was that we would have to create the directory, so this simplifies things by removing that part and trying to save the file in the home directory. We could let people specify this at some point if we want.

sunnyguduru (approved):

PR #189, opened 2024-04-18
feat: track cli command run events

sunnyguduru:
Here's what the properties look like:
CleanShot 2024-04-18 at 16 53 41@2x


cmd/utils/utils.go
```diff
@@ -0,0 +1,47 @@
...
+func (c CommandRunEventType) buildProperties(cmd *cobra.Command) map[string]interface{} {
+	id := uuid.New()
+	baseURI := viper.GetString(cliflags.BaseURIFlag)
+	var flags []string
+	if cmd.HasFlags() {
+		flagSet := cmd.Flags()
+		flagSet.Visit(func(f *pflag.Flag) {
+			flags = append(flags, f.Name)
+		})
+	}
```
sunnyguduru, line 35: managed to dynamically get all the flags used for a command

cmd/utils/utils.go
```diff
@@ -0,0 +1,47 @@
+package utils
```
dbolson, line 1: Calling this `utils` is an anti-pattern since it could become a a dumping ground for random stuff. It seems to make sense to have this in the `analytics` package since it deals with analytics-specific functionality.

cmd/utils/utils.go
```diff
@@ -0,0 +1,47 @@
...
+	"ldcli/cmd/cliflags"
+	"ldcli/internal/analytics"
+
+	"github.com/google/uuid"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+type CommandRunEventType struct {
```
dbolson, line 13: I'm not sure we're getting much from this type instead of passing in the event name directly to `SendEvents`.

cmd/utils/utils.go
dbolson: You could one-liner `cmd.Flags().Visit()` and get rid of the `if cmd.HasFlags()` check.

cmd/utils/utils.go
```diff
@@ -0,0 +1,47 @@
...
+		})
+	}
+
+	properties := map[string]interface{}{
+		"name":   c.EventName,
+		"action": cmd.CalledAs(),
+		"flags":  flags,
+		"id":     id.String(),
+	}
+	if baseURI != "https://app.launchdarkly.com" {
```
dbolson, line 43: Could you extract this to a constant since it's used in multiple places?

cmd/utils/utils.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/spf13/viper"
+)
+
+type CommandRunEventType struct {
+	EventName string
+}
+
+func (c CommandRunEventType) SendEvents(analyticsTracker analytics.Tracker, cmd *cobra.Command) {
+	analyticsTracker.SendEvent(
+		viper.GetString(cliflags.AccessTokenFlag),
```
dbolson, line 19: Can you remove the viper references in here and pass in the values instead?

dbolson (changes requested):
Can you add tests?

cmd/utils/utils.go
sunnyguduru: I wanted to encapsulate the "buildProperties" for each different event type. The other option is to have multiple functions along the lines of BuildPropertiesForCommandRun(), BuildPropertiesForCommandCompleted().
Which I'm open to as well, just that the function name looked quite long.

cmd/environments/get_test.go
```diff
@@ -138,8 +139,46 @@ func TestGet(t *testing.T) {
...
+			"--access-token", "testAccessToken",
+			"--base-uri", "http://test.com",
+			"--environment", "test-env",
+			"--project", "test-proj",
+		}
+
+		_, err := cmd.CallCmd(t, clients, &tracker, args)
+		tracker.AssertCalled(t, "SendEvent", mockedTrackingArgs...)
+		require.NoError(t, err)
+	})
```
sunnyguduru, line 183: Added test for commands

internal/analytics/event_utils.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/google/uuid"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+func CmdRunEventProperties(cmd *cobra.Command, name string) map[string]interface{} {
+	id := uuid.New()
+	baseURI := viper.GetString(cliflags.BaseURIFlag)
```
dbolson, line 14: Can you pass this value in instead of using viper? It removes the dependency.

cmd/projects/list_test.go
```diff
@@ -115,8 +116,43 @@ func TestList(t *testing.T) {
 			"--base-uri", "invalid",
 		}
 
-		_, err := cmd.CallCmd(t, clients, args)
+		_, err := cmd.CallCmd(t, clients, &analytics.NoopClient{}, args)
 
 		assert.EqualError(t, err, "base-uri is invalid"+errorHelp)
 	})
+
+	t.Run("will track analytics for 'CLI Command Run' event", func(t *testing.T) {
```
dbolson, line 124: Can you remove the quotes for these tests? It messes up the test name when running with `-v`.

cmd/environments/environments.go
```diff
@@ -17,12 +19,20 @@ func NewEnvironmentsCmd(
...
 
-	getCmd, err := NewGetCmd(analyticsTracker, client)
+	getCmd, err := NewGetCmd(client)
 	if err != nil {
 		return nil, err
 	}
-
 	cmd.AddCommand(getCmd)
 
+	cmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
```
dbolson, line 28: Can you set this when we create `cmd` above? Otherwise it seems like it's doing something special that wouldn't work if it were set at the same time as the other fields.

internal/analytics/event_utils.go
```diff
@@ -0,0 +1,30 @@
+package analytics
```
dbolson, line 1: Can you rename this file to be `events.go` since "utils" doesn't add any meaning?

cmd/cliflags/baseuri.go
```diff
@@ -0,0 +1,3 @@
+package cliflags
+
+const BaseURIDefault = "https://app.launchdarkly.com"
```
dbolson, line 3: I don't think this needs to be its own file.

cmd/environments/get_test.go
```diff
@@ -138,8 +139,46 @@ func TestGet(t *testing.T) {
...
 
-		_, err := cmd.CallCmd(t, clients, args)
+		_, err := cmd.CallCmd(t, clients, &analytics.NoopClient{}, args)
 
 		assert.EqualError(t, err, "base-uri is invalid"+errorHelp)
 	})
+
+	t.Run("will track analytics for 'CLI Command Run' event", func(t *testing.T) {
+		id := "test-id"
+		mockedTrackingArgs := []interface{}{
```
dbolson, line 149: You could remove some duplication by extracting this to a helper function.
```
// analytics/client.go
func MockedTracker(name string, action string, flags []string) *MockTracker {
	id := "test-id"
	mockedTrackingArgs := []interface{}{
		"testAccessToken",
		"http://test.com",
		"CLI Command Run",
		map[string]interface{}{
			"action":  action,
			"baseURI": "http://test.com",
			"flags":   flags,
			"id":      id,
			"name":    name,
		},
	}
	tracker := MockTracker{ID: id}
	tracker.On("SendEvent", mockedTrackingArgs...)

	return &tracker
}
```
and call it in the tests like
```
tracker := analytics.MockedTracker(
    "flags",
    "get",
    []string{"access-token", "base-uri", "environment", "flag", "project"}, // these need to be alphabetized
)
```

dbolson (changes requested):

cmd/flags/update_test.go
```diff
@@ -285,7 +280,7 @@ func TestToggle(t *testing.T) {
 			"--environment", "test-env-key",
 		}
 
-		_, err := cmd.CallCmd(t, clients, &tracker, args)
+		_, err := cmd.CallCmd(t, clients, tracker, args)
 		tracker.AssertCalled(t, "SendEvent", mockedTrackingArgs...)
```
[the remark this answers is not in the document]
sunnyguduru: Ok, that makes sense. I'll also get rid of mocked args

cmd/config/config.go
```diff
@@ -19,12 +21,20 @@ const (
...
 )
 
-func NewConfigCmd() *cobra.Command {
+func NewConfigCmd(analyticsTracker analytics.Tracker) *cobra.Command {
 	cmd := &cobra.Command{
 		Long:  "View and modify specific configuration values",
 		RunE:  run(),
 		Short: "View and modify specific configuration values",
 		Use:   "config",
+		PreRun: func(cmd *cobra.Command, args []string) {
```
dbolson, line 30: Why is this PreRun while the others are PersistentPreRun?

dbolson (approved):

cmd/config/config.go
sunnyguduru: I considered leaving a comment. PersistentPreRun from my testing only ran on child cmds attached to the parent cmd. The parent cmd is where PersistentPreRun is defined. 

On the other hand PreRun will run on the parent command only. In this case config doesn't have any sub commands similar to the others (flags, environments). Hence why its a `PreRun` and not `PersistentPreRun`

PR #197, opened 2024-04-22
feat: output flag plaintext

dbolson:
Use the output flag to show either JSON or plaintext for a command's response.

This only works for `environments get`, but another PR will add support to the rest of the commands.

**Requirements**

**Related issues**

**Additional context**


cmd/environments/get.go
```diff
@@ -68,6 +69,14 @@ func runGet(
 			return err
 		}
 
+		output, err := output.CmdOutput(
+			viper.GetString(cliflags.OutputFlag),
+			output.SingularOutput(response),
```
dbolson, line 74: We're going to have more types of output, so another PR will add more of these.

internal/output/output.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// resource is the subset of data we need to display a command's plain text response.
+type resource struct {
+	Key  string `json:"key"`
+	Name string `json:"name"`
+}
+
+// CmdOutput returns a command's response as a string formatted based on the user's requested type.
+func CmdOutput(outputKind string, outputter OutputterFn) (string, error) {
```
dbolson, line 44: This is the main function in this package. We use the `OutputterFn` factory so we can do error handling here instead of at the `CmdOutput` call site in the command handler.
```
// with factory fn
output, err := output.CmdOutput(
    viper.GetString(cliflags.OutputFlag),
    output.SingularOutput(response),
)
if err != nil {
    return err
}
```
vs
```
// without factory fn
outputter, err := output.SingularOutput(response)
if err != nil {
    return err
}
output, err := output.CmdOutput(
    viper.GetString(cliflags.OutputFlag),
    outputter
)
```

internal/output/output.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	switch outputKind {
+	case "json":
+		return o.JSON(), nil
+	case "plaintext":
+		return o.String(), nil
+	}
+
+	return "", ErrInvalidOutputKind
```
dbolson, line 57: We should never get here since we validate the input in the command handler.

internal/output/singular_outputter.go
```diff
@@ -0,0 +1,34 @@
+package output
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+var singularPlaintextOutputFn = func(r resource) string {
```
dbolson, line 8: We're going to end up with a few of these functions depending on the shape of the response.

cmd/validators/validators.go
```diff
@@ -48,25 +48,14 @@ func CmdError(err error, commandPath string) error {
 	return errors.New(errorMessage)
 }
 
-func getCommandPath(cmd *cobra.Command) string {
```
dbolson, line 51: Sneaking this in here based on [this comment](https://github.com/launchdarkly/ldcli/pull/195/files#r1575390041).

dbolson:
Moving this to draft to work on it more.

sunnyguduru (approved):

internal/output/output.go
```diff
@@ -0,0 +1,61 @@
...
+	"strings"
+
+	"ldcli/internal/errors"
+)
+
+var ErrInvalidOutputKind = errors.NewError("output is invalid")
+
+// Outputter defines the different ways a command's response can be formatted based on
+// user input.
+type Outputter interface {
```
dbolson, line 13: We need an interface to define how to output JSON and and plain text versions of a response. This will be for a singular resource and when there are multiple resources.

cmd/environments/get.go
```diff
@@ -68,6 +69,14 @@ func runGet(
 			return err
 		}
 
+		output, err := output.CmdOutput(
```
dbolson, line 72: This is the implementation. Another PR will add it for all endpoints and add functionality to work for a response with multiple resources.

PR #200, opened 2024-04-23
feat: added cmd completed to commands

sunnyguduru:
Add CLI Command Completed event for the following:
- Successful run of a subcommand
- API error encountered when running subcommand

Broke out help tracking into its own shortcut ticket.

**Requirements**

**Related issues**

**Additional context**


main.go
```diff
@@ -17,7 +19,7 @@ func main() {
 	httpClient := &http.Client{
 		Timeout: time.Second * 3,
 	}
-	analyticsClient := &analytics.Client{HTTPClient: httpClient}
+	analyticsClient := &analytics.Client{HTTPClient: httpClient, ID: uuid.New().String()}
```
sunnyguduru, line 22: Since a new analyticsClient is created for every new command run. We can just this to our advantage and associate a ID.

internal/analytics/client.go
```diff
@@ -32,6 +33,7 @@ func (c *Client) SendEvent(
 	eventName string,
 	properties map[string]interface{},
 ) {
+	properties["id"] = c.ID
```
sunnyguduru, line 36: In SendEvent we'll intercept  properties and add our id. This means that both CLI Command Run and CLI Command Completed would share the same id.

cmd/environments/environments.go
```diff
@@ -25,6 +25,25 @@ func NewEnvironmentsCmd(
...
+			}
+
+			analyticsTracker.SendEvent(
+				viper.GetString(cliflags.AccessTokenFlag),
+				viper.GetString(cliflags.BaseURIFlag),
+				"CLI Command Completed",
+				map[string]interface{}{
+					"outcome": outcome,
+				},
+			)
```
sunnyguduru, line 45: I'll be abstracting this out into a function. But while I have the first one written, do you want to check it @dbolson

main.go
dbolson: Could we get rid of setting the ID in `CmdRunEventProperties`?

cmd/environments/environments.go
```diff
@@ -25,6 +25,16 @@ func NewEnvironmentsCmd(
 				analytics.CmdRunEventProperties(cmd, "environments"),
 			)
 		},
+		PersistentPostRun: func(cmd *cobra.Command, args []string) {
+			analyticsTracker.SendEvent(
+				viper.GetString(cliflags.AccessTokenFlag),
+				viper.GetString(cliflags.BaseURIFlag),
+				"CLI Command Completed",
```
dbolson, line 32: Seems like we could make separate functions for `CLI Command Run` events and `CLI Command Completed` events so we don't have to type those strings out everywhere. Something like `SendCommandRunEvent()` and `SendCommandCompletedEvent()`.

dbolson:
This seems fine.

main.go
sunnyguduru: Done

cmd/environments/environments.go
sunnyguduru: Yeah that sounds reasonable

internal/analytics/events.go
```diff
@@ -42,8 +46,28 @@ func MockedTracker(name string, action string, flags []string) *MockTracker {
...
+		"CLI Command Completed",
+		map[string]interface{}{
+			"id":      id,
+			"outcome": outcome,
+		},
+	}...)
 	return &tracker
 }
+
+func Outcome(cmd *cobra.Command) string {
```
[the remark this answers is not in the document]
sunnyguduru: Thanks for pointing it out. Forgot to remove it

cmd/environments/environments.go
```diff
@@ -33,5 +26,12 @@ func NewEnvironmentsCmd(
 	}
 	cmd.AddCommand(getCmd)
 
+	for _, c := range cmd.Commands() {
+		c.SetHelpFunc(func(c *cobra.Command, args []string) {
+			analytics.SendCommandRunEvent("environments", c, analyticsTracker)
+			c.Root().Annotations = map[string]string{"help": "true"}
+		})
+	}
+
```
sunnyguduru, line 35: The only way to get "CLI Command Run" when a help command is run is to intercept it here.

cmd/root.go
```diff
@@ -135,6 +135,9 @@ func NewRootCommand(
 }
 
 func Execute(analyticsTracker analytics.Tracker, version string) {
+	outcome := analytics.SUCCESS
+	defer analytics.SendCommandCompletedEvent(&outcome, analyticsTracker)
```
sunnyguduru, line 139: We can then send the Completed tracking event when execute is finishing up

sunnyguduru:
Added tests to only flags get like we discussed in slack DMs.

cmd/cmdtest.go
```diff
@@ -31,7 +31,12 @@ func CallCmd(
 	rootCmd.SetArgs(args)
 
 	err = rootCmd.Execute()
+	outcome := analytics.SUCCESS
```
dbolson, line 34: Do we need this here? I commented it out and the tests still pass.

internal/analytics/events.go
```diff
@@ -42,8 +45,36 @@ func MockedTracker(name string, action string, flags []string) *MockTracker {
...
+		"CLI Command Completed",
+		map[string]interface{}{
+			"id":      id,
+			"outcome": outcome,
+		},
+	}...)
 	return &tracker
 }
+
+func SendCommandRunEvent(name string, cmd *cobra.Command, analyticsTracker Tracker) {
```
dbolson, line 61: How about making these methods on the tracker instead of functions that take a tracker?
```
func (a analyticsTracker) SendCommandRunEvent(accessToken string, baseURI string, eventName string, properties map[string]interface{}) {}
func (a analyticsTracker) SendCommandCompletedEvent(accessToken string, baseURI string, outcome string) {}
```

internal/analytics/events.go
```diff
@@ -42,8 +45,36 @@ func MockedTracker(name string, action string, flags []string) *MockTracker {
...
+	cmd.Root().Annotations = map[string]string{"tracking": "true"}
+	analyticsTracker.SendEvent(
+		viper.GetString(cliflags.AccessTokenFlag),
```
dbolson, line 64: I mentioned this in another PR, but the internal package shouldn't reference framework code like cobra or viper. We should pass in these values instead of reading them from viper. It also means we should move `CmdRunEventProperties` to the `cmd` package (or a subpackage) -- maybe `cmd/analytics`?
* `internal/analytics` - "domain model" for analytics tracking
* `cmd/analytics` - translation layer between framework code and domain code

cmd/root.go
```diff
@@ -152,7 +152,12 @@ func Execute(analyticsTracker analytics.Tracker, version string) {
 	}
 
 	err = rootCmd.Execute()
+	outcome := analytics.SUCCESS
+	if _, tracking := rootCmd.Annotations["tracking"]; tracking {
```
dbolson, line 156: Do we need to set the annotation? Would it would to do this?
```
defer func() {
	analytics.SendCommandCompletedEvent(&outcome, analyticsTracker)
}()
```

internal/analytics/events.go
```diff
@@ -42,8 +45,36 @@ func MockedTracker(name string, action string, flags []string) *MockTracker {
...
+		viper.GetString(cliflags.BaseURIFlag),
+		"CLI Command Run",
+		CmdRunEventProperties(cmd, name),
+	)
+}
+
+func SendCommandCompletedEvent(outcome *string, analyticsTracker Tracker) {
```
dbolson, line 71: Why is `outcome` a pointer to a string instead of a string?

dbolson (changes requested):

cmd/root.go
sunnyguduru: I realized that since we're generating new resources and operations, and we might not have "CLI Command Run" set up to run on all of them. So I'm thinking it doesn't make sense to send the event "CLI Command Completed" unless the previous one does.

sunnyguduru: I wanted to make sure we avoid sending "CLI Command Completed" when "CLI Command Run" hasn't run. 
The reason "CLI Command Completed" always runs is because its in `cmd/root.go`. That was the only way I could get errors to be tracked.

internal/analytics/events.go
sunnyguduru: I changed this. outcome is now just a string

cmd/cmdtest.go
sunnyguduru: Looking into this

sunnyguduru: I realized we don't need this variable anyway. But my main worry from this comment was that the tests weren't working properly. I double checked that the "CLI Command Completed" event must have the right outcome "success" or "error" to for the tests to pass.

cmd/root.go
dbolson: That makes sense since we wouldn't want to send a completed event without a run event. What about setting state on the analytics tracker? Then we could call `SendCommandCompletedEvent` but it would be a no-op if the internal boolean value is false.
```
type Client struct {
	ID         string
	HTTPClient *http.Client
    sentRunEvent bool // new field
	wg         sync.WaitGroup
}

func (c *Client) SendCommandRunEvent() {
    c.sentRunEvent = true
}

func (c *Client) SendCommandCompletedEvent(outcome, accessToken, baseURI string) {
    if c.sentRunEvent {
        // make request
    }
}
```

cmd/config/config.go
```diff
@@ -28,12 +29,11 @@ func NewConfigCmd(analyticsTracker analytics.Tracker) *cobra.Command {
 		RunE:  run(),
 		Short: "View and modify specific configuration values",
 		Use:   "config",
-		PreRun: func(cmd *cobra.Command, args []string) {
-			analyticsTracker.SendEvent(
+		PreRun: func(c *cobra.Command, args []string) {
```
dbolson, line 32: Do we need to change this to c instead of cmd? The latter seems more idiomatic.

cmd/root.go
sunnyguduru: That makes sense. The thought occurred to me too and I definitely like the idea of no-op based on the boolean.

internal/analytics/client.go
```diff
@@ -18,9 +18,20 @@ type Tracker interface {
 		eventName string,
 		properties map[string]interface{},
 	)
+	SendCommandRunEvent(
```
dbolson, line 21: We don't need `SendEvent` in the interface now, and it could be an unexported method in the implementations.

cmd/root.go
sunnyguduru: Done

internal/analytics/client.go
```diff
@@ -83,27 +91,71 @@ func (c *Client) SendEvent(
...
+	properties map[string]interface{},
+) {
+	c.sendEvent(
+		accessToken,
+		baseURI,
+		optOut,
+		"CLI Command Run",
+		properties,
+	)
+	if !optOut {
```
dbolson, line 107: I'm not sure this adds anything since `SendCommandCompletedEvent` will call `c.sendEvent` which will no-op if optOut is true.

dbolson (approved):

internal/analytics/client.go
sunnyguduru: True, I'll address this in a follow up

PR #204, opened 2024-04-24
feat: output flag errors

dbolson:
Adds plaintext support for command error responses.

There's a bigger refactor in here as well since adding in plaintext error handling drew out a simpler implementation than before. A lot of the changes are changes to the call site in each command which adds a lot of +/- changes to the PR.

**Requirements**

**Related issues**

**Additional context**


cmd/config/config.go
```diff
@@ -64,9 +64,10 @@ func run() func(*cobra.Command, []string) error {
 				return err
 			}
 
-			output, err := output.CmdOutput(
+			output, err := output.CmdOutputResource(
```
dbolson, line 67: The output API changed to
1. split into two functions: one for a singular resource response and one for multiple resources
2. pass in the outputter function from the caller because there are as of now five different ways we can get a response to format

We could further refactor by creating a function for each possible combination of singular/multiple response + type of response.

cmd/environments/get.go
```diff
@@ -63,15 +64,25 @@ func runGet(
 			viper.GetString(cliflags.ProjectFlag),
 		)
 		if err != nil {
-			return err
+			output, err := output.CmdOutputResource(
```
dbolson, line 67: We handle all errors the same way.

cmd/flags/get.go
```diff
@@ -57,11 +58,6 @@ func NewGetCmd(client flags.Client) (*cobra.Command, error) {
 
 func runGet(client flags.Client) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
-		// rebind flags used in other subcommands
```
dbolson, line 60: Turns out we don't need to do this. It will be easier to remove these with the generated commands since they'll only be written once.

cmd/members/invite.go
```diff
@@ -61,15 +62,25 @@ func runInvite(client members.Client) func(*cobra.Command, []string) error {
...
+
+			return errors.NewError(output)
 		}
 
-		output, err := output.CmdOutput(
+		output, err := output.CmdOutputResources(
 			viper.GetString(cliflags.OutputFlag),
-			output.NewSingularOutput(response),
+			response,
+			output.MultipleEmailPlaintextOutputFn,
```
dbolson, line 80: Here's an example of a different function used because the response is
```
[
    {
        "email": ""
    }
]
```
which is different from a normal "list" type of response because
1. it doesn't have "items" as a property
2. it has "email" and "_id" instead of "name" and "key"

cmd/projects/create_test.go
```diff
@@ -77,19 +77,15 @@ func TestCreate(t *testing.T) {
...
 		args := []string{
-			"projects",
-			"create",
-			"--access-token",
-			"testAccessToken",
-			"--base-uri",
-			"http://test.com",
-			"-d",
-			`{"key": "test-key", "name": "test-name"}`,
+			"projects", "create",
```
dbolson, line 85: Since these are pairs, it reads a little easier this way.

internal/output/output.go
```diff
@@ -23,7 +23,7 @@ type OutputterFn interface {
 }
 
 // PlaintextOutputFn represents the various ways to output a resource or resources.
-type PlaintextOutputFn[T any] func(t T) string
+type PlaintextOutputFn func(resource) string
```
dbolson, line 26: These only work on one type now.

internal/output/output.go
```diff
@@ -36,13 +36,45 @@ type resources struct {
...
+		resource:     r,
+		resourceJSON: input,
+	})
+}
+
+func CmdOutputResources(outputKind string, input []byte, fn PlaintextOutputFn) (string, error) {
+	var r resources
+	err := json.Unmarshal(input, &r)
+	if err != nil {
+		// sometimes a response doesn't include each item in an "items" property
```
dbolson, line 61: It'd be nice not to have to do this, but turns out our API is a little inconsistent...

internal/output/output.go
```diff
@@ -52,14 +84,3 @@ func CmdOutput(outputKind string, outputter OutputterFn) (string, error) {
 
 	return "", ErrInvalidOutputKind
 }
-
-// FormatColl applies a formatting function to every element in the collection and returns it as a
-// string.
-func formatColl[T any](coll []T, formatFn func(T) string) string {
```
dbolson, line 58: Moved this to be closer to where it's called.

internal/output/outputters.go
```diff
@@ -0,0 +1,42 @@
+package output
+
+import "strings"
+
+type MultipleOutputter struct {
```
dbolson, line 5: I could move the interface definition in here as well if that would make it easier to reason about. I left it in `output.go` to keep that as the one place where we define types that are implemented in other places, but it's not a big deal either way.

internal/output/plaintext_fns.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	return strings.Join(lst, "\n")
+}
+
+// ErrorPlaintextOutputFn converts the resource to plain text specifically for data from the
+// error file.
+// An error response could have a code and message or just a message. It's also possible that
+// there isn't either property.
+var ErrorPlaintextOutputFn = func(r resource) string {
```
dbolson, line 30: This is weird, but hopefully the tests describe what's going on.

cmd/environments/get.go
k3llymariee: maybe we consider renaming this since method/the struct we dump these objects into since they're handling more than just `Resources` in the way that we're defining them elsewhere - e.g. it's also handling errors, config, not just `Resources` like flags, projects, environments, etc. not going to die on that hill though

k3llymariee (approved):
looks great! this feels like it will work better with the generated/templated cmds as well

cmd/environments/get.go
dbolson: Renamed to `CmdOutputSingular` and `CmdOutputMultiple` to be more generic.

PR #206, opened 2024-04-24
feat: allow users to opt out of analytics tracking

k3llymariee:
1. adds a new `analytics-opt-out` flag
a. this value can be set as an environment variable (`LD_ANALYTICS_OPT_OUT`), config value, or passed as a flag
b. this value can be set as `0` or `false` for false, `1` or `true` for true

2. will not send a tracking event if this value is set to true

Manually tested, was unsure how to update tests for this 🤔

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -28,6 +28,9 @@ func NewConfigCmd(analyticsTracker analytics.Tracker) *cobra.Command {
 		Short: "View and modify specific configuration values",
 		Use:   "config",
 		PreRun: func(cmd *cobra.Command, args []string) {
+			if viper.GetBool(cliflags.AnalyticsOptOut) {
+				return
+			}
 			analyticsTracker.SendEvent(
 				viper.GetString(cliflags.AccessTokenFlag),
 				viper.GetString(cliflags.BaseURIFlag),
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -28,12 +28,10 @@ func NewConfigCmd(analyticsTracker analytics.Tracker) *cobra.Command {
 		Short: "View and modify specific configuration values",
 		Use:   "config",
 		PreRun: func(cmd *cobra.Command, args []string) {
-			if viper.GetBool(cliflags.AnalyticsOptOut) {
-				return
-			}
 			analyticsTracker.SendEvent(
 				viper.GetString(cliflags.AccessTokenFlag),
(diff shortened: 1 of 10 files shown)


sunnyguduru (commented):
Can we also make a change somewhere to pass in the noop tracking client based on this flag? I thought in `main.go` but I'm not sure the flags are loaded in yet there.

k3llymariee:
> Can we also make a change somewhere to pass in the noop tracking client based on this flag? I thought in main.go but I'm not sure the flags are loaded in yet there.

@sunnyguduru yeah I'm pretty sure we don't have it there, which is why we add the access token/base URI where/when we do for all our clients. not sure how else we'd do that then unless you have another idea!

sunnyguduru:
> > Can we also make a change somewhere to pass in the noop tracking client based on this flag? I thought in main.go but I'm not sure the flags are loaded in yet there.
> 
> @sunnyguduru yeah I'm pretty sure we don't have it there, which is why we add the access token/base URI where/when we do for all our clients. not sure how else we'd do that then unless you have another idea!

How about in `cmd/root.go`? In the `NewRootCommand` function. I think we can redefine analyticsTracker there in that function to be the no-op version. That function is also where we set the baseUri and access token. So we should be able to access other flags there too

cmd/config/config.go
```diff
@@ -28,6 +28,9 @@ func NewConfigCmd(analyticsTracker analytics.Tracker) *cobra.Command {
 		Short: "View and modify specific configuration values",
 		Use:   "config",
 		PreRun: func(cmd *cobra.Command, args []string) {
+			if viper.GetBool(cliflags.AnalyticsOptOut) {
```
dbolson, line 31: I feel like we could pass this value to the `Send*` methods (since there are going to be a couple after [this PR](https://github.com/launchdarkly/ldcli/pull/200) is merged) and let the analytics tracker do a no-op if it's true. Then we wouldn't need to check every place the `Send*` methods are called.

That said, if it's easier to merge this and then do another PR once the above PR is merged, then go for it.

dbolson (approved):

dbolson:
I'm not sure the flags are fully bound in `NewRootCommand`. You'd have to check a flag value from the config file, an env var, and a flag to see when it's set.

cmd/config/config.go
k3llymariee: easy enough! updated

PR #211, opened 2024-04-25
feat: add hardcoded operation command with no body

k3llymariee:
admittedly this probably didn't need to be its own PR, but we've added logging!

adds `teams get --teamKey {team key}`

note that the flag is camelcase, but that's how it comes from the openapi spec. we'll update this once we do the generation of OperationDate from the OpenAPI spec (comment left)

### DIFF
--- a/cmd/resources/resources.go
+++ b/cmd/resources/resources.go
@@ -17,7 +17,6 @@ func NewResourceCmd(parentCmd *cobra.Command, resourceName, shortDescription, lo
 		Use:   resourceName,
 		Short: shortDescription,
 		Long:  longDescription,
-		//TODO: add tracking here
 	}
 
 	parentCmd.AddCommand(cmd)
@@ -40,6 +39,7 @@ type Param struct {
 	In          string
 	Description string
 	Type        string
+	Required    bool
 }
 
 type OperationCmd struct {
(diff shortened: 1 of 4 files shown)


dbolson (approved):

cmd/resources/resources.go
```diff
@@ -8,16 +8,25 @@ import (
...
 		Long:  longDescription,
-		//TODO: add tracking here
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			analyticsTracker.SendCommandRunEvent(
+				viper.GetString(cliflags.AccessTokenFlag),
+				viper.GetString(cliflags.BaseURIFlag),
+				viper.GetBool(cliflags.AnalyticsOptOut),
+				cmdAnalytics.CmdRunEventProperties(cmd, resourceName),
+			)
+		},
```
k3llymariee, line 29: @sunnyguduru this look about right to you? I tested this and it seems to be working. also yes i did have `resourceName` hardcoded as a string for some reason initially, which is why ben saw those logs 😬

sunnyguduru: Yeah that looks right!

sunnyguduru (approved):

PR #215, opened 2024-04-29
feat: track cli setup started event

sunnyguduru:
Added the `cli setup started` event specifically.
- Created a eventTrackedMsg and trackSetupStartedEvent function to send cmd. 
- All trackSetupStartedEvent function calls happen within container

gif:
CleanShot 2024-04-30 at 11 42 25

**Requirements**

**Related issues**

**Additional context**

### DIFF
--- a/cmd/quickstart.go
+++ b/cmd/quickstart.go
@@ -11,25 +11,28 @@ import (
 
 	"ldcli/cmd/cliflags"
 	"ldcli/cmd/validators"
+	"ldcli/internal/analytics"
 	"ldcli/internal/environments"
 	"ldcli/internal/flags"
 	"ldcli/internal/quickstart"
 )
 
 func NewQuickStartCmd(
+	analyticsTracker analytics.Tracker,
 	environmentsClient environments.Client,
 	flagsClient flags.Client,
 ) *cobra.Command {
 	return &cobra.Command{
 		Args:  validators.Validate(),
 		Long:  "",
-		RunE:  runQuickStart(environmentsClient, flagsClient),
+		RunE:  runQuickStart(analyticsTracker, environmentsClient, flagsClient),
 		Short: "Setup guide to create your first feature flag",
 		Use:   "setup",
 	}
 }
 
 func runQuickStart(
+	analyticsTracker analytics.Tracker,
 	environmentsClient environments.Client,
 	flagsClient flags.Client,
 ) func(*cobra.Command, []string) error {
@@ -42,6 +45,7 @@ func runQuickStart(
 		defer f.Close()
 
 		_, err = tea.NewProgram(quickstart.NewContainerModel(
+			analyticsTracker,
 			environmentsClient,
 			flagsClient,
 			viper.GetString(cliflags.AccessTokenFlag),
--- a/cmd/quickstart.go
+++ b/cmd/quickstart.go
@@ -45,6 +45,7 @@ func runQuickStart(
 		defer f.Close()
 
 		_, err = tea.NewProgram(quickstart.NewContainerModel(
+			viper.GetBool(cliflags.AnalyticsOptOut),
 			analyticsTracker,
 			environmentsClient,
 			flagsClient,
--- a/cmd/quickstart.go
+++ b/cmd/quickstart.go
@@ -45,11 +45,11 @@ func runQuickStart(
 		defer f.Close()
 
 		_, err = tea.NewProgram(quickstart.NewContainerModel(
-			viper.GetBool(cliflags.AnalyticsOptOut),
 			analyticsTracker,
 			environmentsClient,
 			flagsClient,
 			viper.GetString(cliflags.AccessTokenFlag),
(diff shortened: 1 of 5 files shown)


internal/quickstart/messages.go
```diff
@@ -240,3 +241,13 @@ func selectedSDK(index int) tea.Cmd {
...
+
+type eventTrackedMsg struct{}
+
+func trackSetupStartedEvent(tracker analytics.Tracker, accessToken, baseURI string, optOut bool, step string) tea.Cmd {
+	return func() tea.Msg {
+		tracker.SendSetupStartedEvent(accessToken, baseURI, optOut, step)
+
+		return eventTrackedMsg{}
+	}
+}
```
sunnyguduru, line 253: It's a bit confusing but according to the cli tracking spec. The event "CLI Setup Started" should be called for every step. The Step string will indicate what step we're on.

internal/quickstart/container.go
```diff
@@ -18,21 +21,28 @@ const (
...
-	_ = iota
+	_ step = iota
 	stepCreateFlag
 	stepChooseSDK
 	stepShowSDKInstructions
 	stepToggleFlag
 )
 
+func (s step) String() string {
+	return [...]string{"_", "1 - feature flag name", "2 - sdk selection", "3 - sdk installation", "4 - flag toggle"}[s]
```
dbolson, line 35: Since this slice will always be small, we don't need to set the length of the backing array, so you could write this as
```
func (s step) String() string {
	return []string{
		"_",
		"1 - feature flag name",
		"2 - sdk selection",
		"3 - sdk installation",
		"4 - flag toggle",
	}[s]
}
```
for more readability.

internal/quickstart/container.go
```diff
@@ -46,6 +56,7 @@ type ContainerModel struct {
 }
 
 func NewContainerModel(
+	analyticsTracker analytics.Tracker,
```
dbolson, line 59: It looks like we don't set this on the container model, so it panics later when calling a method on it.

internal/quickstart/container.go
```diff
@@ -64,11 +75,18 @@ func NewContainerModel(
 }
 
 func (m ContainerModel) Init() tea.Cmd {
-	return nil
+	return trackSetupStartedEvent(
+		m.analyticsTracker,
+		m.accessToken,
+		m.baseUri,
+		viper.GetBool(cliflags.AnalyticsOptOut),
```
dbolson, line 82: Since this is in `internal`, it shouldn't call `viper` directly.

internal/quickstart/messages.go
dbolson: It would be worth clarifying the event name for subsequent steps since it is confusing.

dbolson (changes requested):

internal/quickstart/messages.go
sunnyguduru: @bwoskow-ld Does this sound right? From the spec it looked like "CLI Setup Started" was what would be used on every step. Just with the step string itself denoting what step we're on.

internal/quickstart/container.go
sunnyguduru: Sounds good

internal/quickstart/messages.go
bwoskow-ld: Oh yeah that is confusing. Let's rename it to `CLI Setup Step Started` since that's more in line with what this data represents. I'll update the spec page.

sunnyguduru:
CleanShot 2024-04-30 at 11 42 25

cmd/quickstart.go
```diff
@@ -42,6 +45,8 @@ func runQuickStart(
 		defer f.Close()
 
 		_, err = tea.NewProgram(quickstart.NewContainerModel(
+			viper.GetBool(cliflags.AnalyticsOptOut),
```
dbolson, line 48: Seems like this should be with the other flag values at the end of the param list.

internal/analytics/client.go
```diff
@@ -128,6 +134,23 @@ func (c *Client) SendCommandCompletedEvent(
 	}
 }
 
+func (c *Client) SendSetupStartedEvent(
```
dbolson, line 137: This should be `SendSetupStepStartedEvent`

internal/quickstart/container.go
```diff
@@ -18,21 +19,35 @@ const (
...
+		"4 - flag toggle",
+	}[s]
+}
+
 // ContainerModel is a high level container model that controls the nested models where each
 // represents a step in the quick-start flow.
 type ContainerModel struct {
 	accessToken        string
+	analyticsTracker   analytics.Tracker
+	analyticsOptOut    bool
```
dbolson, line 47: Can you alphabetize this list?

dbolson (approved):

cmd/quickstart.go
sunnyguduru: Fair. I was putting it on top of analyticsTracker for alphabetical reasons

PR #218, opened 2024-04-29
feat: generic api request function

k3llymariee:
Adds a new client and method to make generic API requests based on the openapi operation structs we'll be creating for each endpoint.


cmd/resource_cmds_test.go
```diff
@@ -1,42 +0,0 @@
-package cmd_test
```
k3llymariee, line 1: file was changed & moved not deleted

cmd/resources/resource_cmds.go
```diff
@@ -1,19 +1,18 @@
 // this file WILL be generated (sc-241153)
 
-package cmd
```
k3llymariee, line 3: changes here are just due to moving the files

internal/resources/client.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"ldcli/internal/errors"
+)
+
+type Client interface {
```
k3llymariee, line 14: made a new interface for testing but had an issue with passing in data to the mock client, will try to tackle in follow up PR

cmd/resources/resources.go
```diff
@@ -97,41 +113,81 @@ func (op *OperationCmd) initFlags() error {
...
+	var urlParms []string
 	for _, p := range op.Params {
-		var val interface{}
-		switch p.Type {
-		case "string":
-			val = viper.GetString(p.Name)
-		case "boolean":
-			val = viper.GetBool(p.Name)
-		case "int":
-			val = viper.GetInt(p.Name)
```
k3llymariee, line 121: i think it's actually ok to just grab all of these as string since we're just passing them into the query param or url anyway

dbolson (approved):

cmd/resources/resources.go
```diff
@@ -97,41 +112,81 @@ func (op *OperationCmd) initFlags() error {
 	return nil
 }
 
-func (op *OperationCmd) makeRequest(cmd *cobra.Command, args []string) error {
-	paramVals := map[string]interface{}{}
+func formatURL(baseURI, path string, urlParams []string) string {
```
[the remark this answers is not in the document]
k3llymariee: changed to `buildURLWithParams`

PR #226, opened 2024-05-01
feat: generate teams operation data from openapi spec

k3llymariee:
Adds a function that reads and parses an OpenAPI spec into our desired resources/operations data structs - for now it works with just the teams data (added that json temporarily at `ld-teams-openapi.json` to run locally).  

Note: this new method isn't actually being called anywhere yet, but you can test this locally with the file above ☝️


cmd/resources/resources.go
```diff
@@ -34,24 +34,6 @@ func NewResourceCmd(parentCmd *cobra.Command, analyticsTracker analytics.Tracker
...
-}
-
-type Param struct {
-	Name        string
-	In          string
-	Description string
-	Type        string
-	Required    bool
-}
-
```
k3llymariee, line 54: moved to `cmd/resources/gen_resources.go` since that's where these are added

cmd/resources/gen_resources_test.go
```diff
@@ -0,0 +1,51 @@
...
+	assert.NoError(t, err)
+
+	expectedFromFile, err := os.ReadFile("test_data/expected_template_data.json")
+	require.NoError(t, err)
+
+	var expected resources.TemplateData
+	err = json.Unmarshal(expectedFromFile, &expected)
+	require.NoError(t, err)
+
+	t.Run("succeeds with single get resource", func(t *testing.T) {
```
k3llymariee, line 24: not running the comparison directly on each op led to flakey tests - if there's a better way to do this LMK!

cmd/resources/gen_resources.go
```diff
@@ -0,0 +1,138 @@
...
+
+type Param struct {
+	Name        string
+	In          string
+	Description string
+	Type        string
+	Required    bool
+}
+
+func GetTemplateData(fileName string) (TemplateData, error) {
```
k3llymariee, line 40: this isn't actually being called anywhere but the tests yet, but I was hacking into `main.go` and passing in the file name to test it out locally. will work on actually calling this in a generate file in a follow up PR.

cmd/resources/gen_resources.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+type TemplateData struct {
+	Resources map[string]*ResourceData
+}
+
+type ResourceData struct {
+	Name        string
+	Description string
+	Operations  map[string]*OperationData
```
k3llymariee, line 19: it's probably not strictly necessary to have these as a `map[string]*OperationData` vs `[]*OperationData` but it was helpful for pulling out a specific op type for testing.

cmd/resources/gen_resources_test.go
dbolson: I'm not getting test failures when running `go test ./cmd/resources/ -run TestGetTemplateData -count=100` after changing the test to
```
t.Run("succeeds with single get resource", func(t *testing.T) {
	assert.Equal(t, expected, actual)
})
```
What failures were you seeing?

cmd/resources/gen_resources.go
```diff
@@ -0,0 +1,138 @@
...
+				schema = spec.Components.Schemas[schemaName]
+			}
+		}
+	}
+
+	if schema == nil {
+		// probably won't need to keep this logging in but leaving it for debugging purposes
+		log.Printf("No response type defined for %s", op.OperationID)
+	} else {
+		for propName, _ := range schema.Value.Properties {
```
dbolson, line 130: I get a lint warning that this could be simplified to `for propName := range`.

dbolson (approved):

cmd/resources/gen_resources_test.go
k3llymariee: ...I swear I had some. But you're right, updated.

cmd/resources/gen_resources.go
```diff
@@ -0,0 +1,138 @@
...
+	"log"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+type TemplateData struct {
+	Resources map[string]*ResourceData
```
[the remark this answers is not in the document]
k3llymariee: yup!

PR #237, opened 2024-05-02
chore: refactor and add build tags to eventually generate templates

k3llymariee:
Breaking down my mega branch into (hopefully) smaller PRs - this just (1) moves around the data structs so we can use them when generating the template data, and (2) adds the build/generate tags we'll be using to create the template and its data.


cmd/resources/gen_resources.go
```diff
@@ -1,139 +1,20 @@
-package resources
+//go:build gen_resources
+// +build gen_resources
```
k3llymariee, line 2: the only real addition (pt 1 of 2)

cmd/root.go
```diff
@@ -1,3 +1,5 @@
+//go:generate go run resources/gen_resources.go
+
```
k3llymariee, line 2: the only real addition (pt 2 of 2)

dbolson (approved):

PR #238, opened 2024-05-03
feat: create and generate template from template data for teams

k3llymariee:
Continuing to work just with the teams data in the OpenAPI spec, generate a template based on the template data to add resource commands to the root command. 

Note that the subcommands are based off of the spec's `operationId` (e.g. `get-team` and `post-team` vs `get` and `create`) and the flags for each subcommand are based off the param values (e.g. `team-key` and `project-key` vs `team` and `project`). I plan to update this is in the next PR where we add the remaining resources from the openapi spec. 

running `teams --help`
image

running `teams get-team --help`
image

checking that required flags are set:
image


cmd/resources/resources.go
```diff
@@ -117,39 +138,42 @@ func GetTemplateData(fileName string) (TemplateData, error) {
...
-			if propName == "items" {
-				use = "list"
-				break
-			}
-		}
-	}
-	return use
+	return strcase.ToKebab(op.OperationID)
+
+	// TODO: work with operation ID & response type to stripe out resource name and update post -> create, get -> list, etc.
```
k3llymariee, line 143: see comment in main description - going to bring this back in but for now, to remove duplicate subcommands, using just the OperationId

cmd/resources/resources.go
```diff
@@ -52,6 +53,11 @@ type Param struct {
 	Required    bool
 }
 
+func jsonString(s string) string {
+	bs, _ := json.Marshal(s)
+	return string(bs)
+}
+
```
k3llymariee, line 60: we have all kinds of markdown stuff in the descriptions of resources and endpoints that was causing the template to throw errors - marshaling to json and returning that string to help!

cmd/resources/resources.go
```diff
@@ -29,6 +29,7 @@ type TemplateData struct {
 
 type ResourceData struct {
 	Name        string
+	DisplayName string
```
k3llymariee, line 32: I can also pass in functions to the template to handle converting these values from camel/kebab, which I might do in a refactor

cmd/resources/resources.go
```diff
@@ -7,10 +7,10 @@ import (
 	"net/url"
 	"os"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/iancoleman/strcase"
```
k3llymariee, line 13: hoping it's ok to use this (..stripe does 🙃)

cmd/resources/resource_cmds.tmpl
```diff
@@ -0,0 +1,43 @@
...
+	gen_{{ $resName }}ResourceCmd := NewResourceCmd(
+        rootCmd,
+        analyticsTracker,
+        "{{ $resData.Name }}",
+        "Make requests (list, create, etc.) on {{ $resData.DisplayName }}",
+        {{ $resData.Description }},
+    )
+    {{ end }}
+
+    // Operation commands{{ range $resName, $resData := .Resources }}{{ range $opName, $opData := $resData.Operations }}
```
dbolson, line 23: Could you break these range calls out into separate lines to more easily see the nesting?

dbolson (approved):

cmd/resources/resource_cmds.tmpl
k3llymariee: yeah it'll add an extra line between then comment and the following commands which bothers me only very slightly 🙃

PR #249, opened 2024-05-07
fix: remove old members cmd

k3llymariee:
remove the old members command before the next release


dbolson (approved):
Can we delete the member command's code too?

PR #250, opened 2024-05-08
chore: Refactor tracker

dbolson:
Refactors the analytics tracker so we don't need to use the access token and base URI values for every method call.

**Requirements**

**Related issues**

**Additional context**


cmd/cmdtest.go
```diff
@@ -22,37 +20,29 @@ var StubbedSuccessResponse = `{
 func CallCmd(
 	t *testing.T,
 	clients APIClients,
-	tracker analytics.Tracker,
+	trackerFn analytics.TrackerFn,
```
dbolson, line 23: We now pass a factory function instead of the tracker itself.

cmd/config/config.go
```diff
@@ -24,7 +24,20 @@ const (
 	UnsetFlag = "unset"
 )
 
-func NewConfigCmd(analyticsTracker analytics.Tracker) *cobra.Command {
+type ConfigCmd struct {
```
dbolson, line 27: We need to carry the state to know if we showed the help text, whether because the user passed `--help` or because they didn't call the command correctly. For non-resource commands, we need to create a struct to hold this state, along with the command itself.

cmd/config/config.go
```diff
@@ -33,18 +46,22 @@ func NewConfigCmd(analyticsTracker analytics.Tracker) *cobra.Command {
 		PreRun: func(cmd *cobra.Command, args []string) {
 			// only track event if there are flags
 			if len(os.Args[1:]) > 1 {
-				analyticsTracker.SendCommandRunEvent(
+				analyticsTrackerFn(
```
dbolson, line 49: Since we don't get the flag values until we run the command, we can pass in the factory function in the command's constructor and call it when we have the flag values. The factory function already has an ID, so additional calls from the same factory function will all have the same ID for correlation.

cmd/config/config.go
```diff
@@ -79,7 +96,7 @@ func NewConfigCmd(analyticsTracker analytics.Tracker) *cobra.Command {
 	cmd.Flags().String(UnsetFlag, "", "Unset a config field")
 	_ = viper.BindPFlag(UnsetFlag, cmd.Flags().Lookup(UnsetFlag))
 
-	return cmd
+	return &configCmd
```
dbolson, line 99: We need a pointer to the command since we're mutating its internal state.

cmd/resources/resource_cmds.go
```diff
@@ -9,228 +9,232 @@ import (
 	"ldcli/internal/resources"
 )
 
-func AddAllResourceCmds(rootCmd *cobra.Command, client resources.Client, analyticsTracker analytics.Tracker) {
+func AddAllResourceCmds(
```
dbolson, line 12: I updated the template and regenerated this file.

cmd/root.go
```diff
@@ -33,12 +33,41 @@ type APIClients struct {
 	ResourcesClient    resources.Client
 }
 
+type Command interface {
```
dbolson, line 36: We need an interface so the root command can keep track of its children that know if we showed a command or action's help. We only need this for non-resource commands because we can handle tracking the help in the root.

cmd/root.go
```diff
@@ -65,15 +94,21 @@ func NewRootCommand(
 		SilenceErrors: true,
 		SilenceUsage:  true,
 	}
+	rootCmd := &RootCmd{
+		cmd: cmd,
+	}
 
 	hf := cmd.HelpFunc()
 	cmd.SetHelpFunc(func(c *cobra.Command, args []string) {
+		rootCmd.helpCalled = true
```
dbolson, line 103: This works for resource commands.

cmd/root.go
```diff
@@ -165,29 +205,34 @@ func Execute(analyticsTracker analytics.Tracker, version string) {
...
-		rootCmd.AddCommand(completionCmd)
+See each command's help for details on how to use the generated script.`, rootCmd.Cmd().Name())
+		rootCmd.Cmd().AddCommand(completionCmd)
 	}
 
 	err = rootCmd.Execute()
-	outcome := analytics.SUCCESS
-	if err != nil {
+
+	var outcome string
```
dbolson, line 218: Refactored this to work with the additional case of when help is called.

cmd/root.go
```diff
@@ -165,29 +205,34 @@ func Execute(analyticsTracker analytics.Tracker, version string) {
...
-	optOut, _ := strconv.ParseBool(optOutStr)
-	analyticsTracker.SendCommandCompletedEvent(
-		rootCmd.PersistentFlags().Lookup(cliflags.AccessTokenFlag).Value.String(),
-		rootCmd.PersistentFlags().Lookup(cliflags.BaseURIFlag).Value.String(),
-		optOut,
-		outcome,
+	analyticsClient := trackerFn.Tracker(version)(
+		rootCmd.cmd.PersistentFlags().Lookup(cliflags.AccessTokenFlag).Value.String(),
+		rootCmd.cmd.PersistentFlags().Lookup(cliflags.BaseURIFlag).Value.String(),
+		viper.GetBool(cliflags.AnalyticsOptOut),
```
dbolson, line 232: I double-checked that this was already set, and it is, so we don't need to get it the command's flags.

cmd/root.go
```diff
@@ -165,29 +205,34 @@ func Execute(analyticsTracker analytics.Tracker, version string) {
...
-		rootCmd.PersistentFlags().Lookup(cliflags.BaseURIFlag).Value.String(),
-		optOut,
-		outcome,
+	analyticsClient := trackerFn.Tracker(version)(
+		viper.GetString(cliflags.AccessTokenFlag),
+		viper.GetString(cliflags.BaseURIFlag),
+		viper.GetBool(cliflags.AnalyticsOptOut),
 	)
+	analyticsClient.SendCommandCompletedEvent(outcome)
+	analyticsClient.Wait()
```
dbolson, line 235: We can move this in here so we don't need to create the tracker factory in main and pass it in.

internal/analytics/client.go
```diff
@@ -7,65 +7,64 @@ import (
...
 
+type TrackerFn func(accessToken string, baseURI string, optOut bool) Tracker
+
+type ClientFn struct {
+	ID string
+}
+
+func (fn ClientFn) Tracker(version string) TrackerFn {
+	return func(accessToken string, baseURI string, optOut bool) Tracker {
+		if optOut {
```
dbolson, line 23: This encapsulates the opt-out flag so we don't need to check it when calling the send methods. If the user opts-out, all the send methods are no-ops.

internal/analytics/client.go
```diff
@@ -7,66 +7,64 @@ import (
...
 }
 
 type Client struct {
-	ID            string
-	HTTPClient    *http.Client
-	Version       string
-	sentHelpEvent bool
-	sentRunEvent  bool
-	wg            sync.WaitGroup
+	accessToken string
```
dbolson, line 57: Added the access token and base URI and unexported the fields.

internal/analytics/client.go
```diff
@@ -7,66 +7,64 @@ import (
...
+	SendSetupSDKSelectedEvent(sdk string)
+	SendSetupFlagToggledEvent(on bool, count int, duration_ms int64)
+	Wait()
 }
 
 type Client struct {
-	ID            string
-	HTTPClient    *http.Client
-	Version       string
-	sentHelpEvent bool
```
dbolson, line 53: These state fields go away since we've pushed the logic out.

internal/quickstart/container.go
```diff
@@ -43,18 +43,17 @@ func (s step) String() string {
 // represents a step in the quick-start flow.
 type ContainerModel struct {
 	accessToken        string
-	analyticsOptOut    bool
 	analyticsTracker   analytics.Tracker
-	baseUri            string
+	baseURI            string
```
dbolson, line 47: We still need the access token and base URI for the LD clients.

k3llymariee (approved):

PR #253, opened 2024-05-08
fix: fix go sdk instructions

k3llymariee:
tried actually setting these up for the next demo day recording and noticed these were off. will let the quickstart team know also

### DIFF
--- a/internal/sdks/sdk_instructions/go.md
+++ b/internal/sdks/sdk_instructions/go.md
@@ -99,8 +99,7 @@ func main() {
 
 Now that your application is ready, run the application to see what value we get.
 ```shell
-export LAUNCHDARKLY_SDK_KEY=1234567890abcdef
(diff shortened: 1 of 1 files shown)


internal/sdks/sdk_instructions/go.md
```diff
@@ -99,9 +99,10 @@ func main() {
 
 Now that your application is ready, run the application to see what value we get.
 ```shell
+export LAUNCHDARKLY_SDK_KEY=1234567890abcdef
```
k3llymariee, line 102: the file gets the environment variable but we didn't actually set it anywhere. this could probably go at the top or here, doesn't matter much where

dbolson (approved):
[Looks like](https://ld-stg.launchdarkly.com/toggle-runner-demo/production/get-started/quickstart) I missed these changes.

PR #256, opened 2024-05-08
fix: set dynamic viewport height/width based on window size

k3llymariee:
depending on your terminal window size the header text wasn't always showing properly because we had a fixed height (and width) for the viewport on the show sdk step. it will now be set when the model is created, and also updated when the window size changes.

### DIFF
--- a/internal/quickstart/show_sdk_instructions.go
+++ b/internal/quickstart/show_sdk_instructions.go
@@ -52,11 +52,13 @@ func NewShowSDKInstructionsModel(
 	url string,
 	flagKey string,
 	environment *environment,
+	height int,
+	width int,
 ) tea.Model {
 	s := spinner.New()
 	s.Spinner = spinner.Points
 
-	vp := viewport.New(viewportWidth, viewportHeight)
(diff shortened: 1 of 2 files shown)


internal/quickstart/show_sdk_instructions.go
```diff
@@ -148,23 +174,13 @@ func (m showSDKInstructionsModel) View() string {
...
-using the %s SDK in your Default project & Test environment.
-
-You should have everything you need to get started, including the flag from
-the previous step and your environmnet key from your Test environment already
-embedded in the code!
-
-Open a new terminal window to get started.
+	headerHeight := lipgloss.Height(m.headerView())
+	footerHeight := lipgloss.Height(m.footerView())
+	containerHeight := lipgloss.Height("\nStep %d of %d\n")
```
k3llymariee, line 179: extra buffer for the text we set at the container level

dbolson (approved):

k3llymariee:
(this is causing the sdk instructions to get cut off in the viewport, not merging until i can get that fixed)

PR #258, opened 2024-05-09
chore: add usage template new resources command

k3llymariee:
Slims down the list of commands we show when running `ldcli --help`, and adds a new command to view the list of all available resource commands. 

Welcoming feedback on:
- which resources we show above-the-line and how we describe them
- want we want to call the ✨special commands✨ like config, completion, setup
- also anything else 😅 

Before:
image

After:
image

running
```shell
ldcli resources help
```
image


cmd/templates.go
```diff
@@ -0,0 +1,23 @@
...
+  {{rpad "setup" 29}} Setup guide to create your first feature flag
+  {{rpad "config" 29}} View and modify specific configuration values
+  {{rpad "completion" 29}} Generate the autocompletion script for the specified shell
+
+Resource Commands:
+  {{rpad "flags" 29}} List, create, modify, and delete feature flags, their statuses, and their expiring targets
+  {{rpad "environments" 29}} Create, delete, and update environments
+  {{rpad "projects" 29}} Create, destroy, and manage projects
+  {{rpad "members" 29}} Invite new members to an account
+  {{rpad "segments" 29}} List, create, modify, and delete segments
```
k3llymariee, line 17: these are snippets pulled out from the tag description in the openapi spec for each of these resources

cmd/resources/resources.go
```diff
@@ -169,6 +190,7 @@ func NewResourceCmd(
 	}
 
 	parentCmd.AddCommand(cmd)
+	parentCmd.Annotations[resourceName] = "resource"
```
k3llymariee, line 193: adding these annotations will allow us to pull each resource from the parent command when running `resources`

cmd/templates.go
```diff
@@ -0,0 +1,23 @@
...
+func getUsageTemplate() string {
+	return `Usage:
+  {{.CommandPath}} [command]
+
+Commands:
+  {{rpad "setup" 29}} Setup guide to create your first feature flag
+  {{rpad "config" 29}} View and modify specific configuration values
+  {{rpad "completion" 29}} Generate the autocompletion script for the specified shell
+
+Resource Commands:
```
mmrj, line 12:
```suggestion
Common resource commands:
```

cmd/templates.go
```diff
@@ -0,0 +1,23 @@
+package cmd
+
+func getUsageTemplate() string {
+	return `Usage:
+  {{.CommandPath}} [command]
+
+Commands:
+  {{rpad "setup" 29}} Setup guide to create your first feature flag
+  {{rpad "config" 29}} View and modify specific configuration values
+  {{rpad "completion" 29}} Generate the autocompletion script for the specified shell
```
mmrj, line 10: Dumb question, but I don't understand what this one means?

Will most ldcli users know what the autocompletion script does / why they'd want one? if yes, this is probably fine. if unsure, maybe something like "Generate a script to [some use/benefit] for the specified shell"

cmd/templates.go
```diff
@@ -0,0 +1,23 @@
+package cmd
+
+func getUsageTemplate() string {
+	return `Usage:
+  {{.CommandPath}} [command]
+
+Commands:
+  {{rpad "setup" 29}} Setup guide to create your first feature flag
```
mmrj, line 8:
```suggestion
  {{rpad "setup" 29}} Create your first feature flag using a step-by-step guide
```
just a suggestion - wanted to start this with a verb to match all the other commands

cmd/templates.go
mmrj: Well, this is embarassingly inconsistent 🙈 sorry about that! I can get this updated in the spec. In the meantime, couple of suggestions:

```suggestion
  {{rpad "flags" 29}} List, create, and modify feature flags and their targeting
  {{rpad "environments" 29}} List, create, and manage environments
  {{rpad "projects" 29}} List, create, and manage projects
  {{rpad "members" 29}} Invite new members to an account
  {{rpad "segments" 29}} List, create, modify, and delete segments
```
reasoning: 
- flags: flag status is very de-emphasized in new shell; you don't usually delete them (usually you archive); "modify" (the semantic patch) has most of its semantic patch instructions around variations and targeting
- environments, projects: use "manage" because you typically don't delete these either (and also so that not every description starts "list, create, modify, and delete" -- will be harder to read)

mmrj (approved):
Thanks @k3llymariee ! added a few suggestions

cmd/templates.go
k3llymariee: 👍 makes sense to me!

cmd/templates.go
k3llymariee: yeah i think so - this is a built in command from the CLI library we're using

mmrj: sounds good, thanks

dbolson (approved):

PR #260, opened 2024-05-09
chore: add more overrides to cmd uses

k3llymariee:
see [gsheet](https://docs.google.com/spreadsheets/d/1VDiDi8sb4MfySWNJzl8rjm1uOGdCrvNBUimIJfYzCZY/edit#gid=1549970466) for mapping of openapi tags & operationIds -> ldcli resources & subcommands

There are still more that we _might_ want to update. I'm getting a little wary the more we move away from the openAPI spec / feels like that probably does need updating at some point.


cmd/resources/resource_utils.go
```diff
@@ -33,6 +34,48 @@ func getResourceNames(name string) (string, string) {
 	return name, resourceKey
 }
 
+var mapOperationIdToCmdUse = map[string]string{
+	"getAuditLogEntries":               "list",
```
dbolson, line 38: Could you alphabetize this?

dbolson (approved):

PR #262, opened 2024-05-10
feat: break out required and optional flags in subcommand help

k3llymariee:
```shell
ldcli teams get --help
```
image

### DIFF
--- a/cmd/resources/resources.go
+++ b/cmd/resources/resources.go
@@ -135,6 +135,9 @@ func GetTemplateData(fileName string) (TemplateData, error) {
 			for _, p := range op.Parameters {
 				if p.Value != nil {
 					// TODO: confirm if we only have one type per param b/c somehow this is a slice
+					if strings.Contains(p.Value.Description, "Deprecated") {
+						continue
+					}
 					types := *p.Value.Schema.Value.Type
 					param := Param{
 						Name:        strcase.ToKebab(p.Value.Name),
@@ -253,6 +256,7 @@ func (op *OperationCmd) initFlags() error {
 			if err != nil {
 				return err
 			}
+			op.cmd.Flags().SetAnnotation(flagName, "required", []string{"true"})
 		}
 
 		err := viper.BindPFlag(flagName, op.cmd.Flags().Lookup(flagName))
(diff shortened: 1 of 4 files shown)


cmd/resources/resources.go
```diff
@@ -135,6 +135,9 @@ func GetTemplateData(fileName string) (TemplateData, error) {
 			for _, p := range op.Parameters {
 				if p.Value != nil {
 					// TODO: confirm if we only have one type per param b/c somehow this is a slice
+					if strings.Contains(p.Value.Description, "Deprecated") {
+						continue
+					}
```
k3llymariee, line 140: sneaking this in here as i noticed we had some deprecated flags in the cli

cmd/resources/resources.go
```diff
@@ -352,10 +356,49 @@ func NewOperationCmd(parentCmd *cobra.Command, client resources.Client, op Opera
...
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
+
+Required flags:
+{{WrappedRequiredFlagUsages . | trimTrailingWhitespaces}}
+
+Optional flags:
+{{WrappedOptionalFlagUsages . | trimTrailingWhitespaces}}
+
+Global Flags:
+{{rpad "  -h, --help" 29}} Help for this command
```
k3llymariee, line 396: it felt weird to include this within the "optional" flags, even though it's not technically a global flag it kinda is? lmk if we want to move this somewhere else

cmd/templates.go
```diff
@@ -21,3 +27,44 @@ Flags:
...
+	width, _, err := term.GetSize(0)
+	if err != nil {
+		width = 80
+	}
+
+	return width
+}
+
+func init() {
+	cobra.AddTemplateFunc("WrappedRequiredFlagUsages", WrappedRequiredFlagUsages)
```
dbolson, line 68: Could you move these to [`Execute()`](https://github.com/launchdarkly/ldcli/blob/main/cmd/root.go#L190) so we don't use `init()`?

dbolson (approved):

cmd/templates.go
k3llymariee: oh yup!

PR #270, opened 2024-05-10
fix: update flag to use clientside id based on selected sdk

k3llymariee:
we'll now send a PATCH request to update the flag if/when the user chooses a client-side sdk. flags created with the go client already support mobile keys [by default](https://github.com/launchdarkly/api-client-go/blob/ea6705ab8a49db1755fdcc47ad60e04cec85eed9/model_client_side_availability_post.go#L19-L24).


internal/quickstart/show_sdk_instructions.go
```diff
@@ -98,6 +105,10 @@ func (m showSDKInstructionsModel) Init() tea.Cmd {
 		cmds = append(cmds, fetchEnv(m.environmentsClient, m.accessToken, m.baseUri, defaultEnvKey, defaultProjKey))
 	}
 
+	if m.sdkKind == clientSideSDK {
```
k3llymariee, line 108: interestingly all the flags we create using the go sdk [have mobile keys enabled by default](https://github.com/launchdarkly/api-client-go/blob/ea6705ab8a49db1755fdcc47ad60e04cec85eed9/model_client_side_availability_post.go#L19-L24)

dbolson (approved):

PR #273, opened 2024-05-10
fix: update usage template to only show flags when present

k3llymariee:
updates the operation usage template to only show a header for required/optional flags if they're actually there. also removes the template for resources in favor of this singular template now. 

truly don't know if this is the Best Way to handle this as I'm not familiar with templates, but it works!

before:
image

after:
image

and also resources work now too, though to be honest I don't know how `Flags:` is showing up there 🤔 
image

### DIFF
--- a/cmd/resources/resources.go
+++ b/cmd/resources/resources.go
@@ -210,7 +210,7 @@ func NewResourceCmd(
 		},
 	}
 
-	cmd.SetUsageTemplate(getResourceUsageTemplate())
+	cmd.SetUsageTemplate(OperationUsageTemplate())
 	parentCmd.AddCommand(cmd)
 	parentCmd.Annotations[resourceName] = "resource"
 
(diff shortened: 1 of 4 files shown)


cmd/resources/resources.go
```diff
@@ -210,7 +210,7 @@ func NewResourceCmd(
 		},
 	}
 
-	cmd.SetUsageTemplate(getResourceUsageTemplate())
+	cmd.SetUsageTemplate(OperationUsageTemplate())
```
dbolson, line 213: Is this still an accurate name, or should we rename it to not be resource- or operation-specific?

dbolson (approved):

cmd/resources/resources.go
k3llymariee: yeah maybe `SubcommandUsageTemplate`?

PR #275, opened 2024-05-13
fix: remove bad template func

k3llymariee:
broke the template with my last commit


cmd/resources/resources.go
```diff
@@ -384,7 +384,7 @@ Available Commands:{{range $cmds}}{{if (or .IsAvailableCommand (eq .Name "help")
   {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if not .AllChildCommandsHaveGroup}}
 
 Additional Commands:{{range $cmds}}{{if (and (eq .GroupID "") (or .IsAvailableCommand (eq .Name "help")))}}
-  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}{{if gt (len WrappedRequiredFlagUsages .) 0}}
```
k3llymariee, line 387: this was the issue

cmd/pager.go
```diff
@@ -0,0 +1,13 @@
+package cmd
```
dbolson, line 1: Do we need this file?

k3llymariee: nooope that was me moving too fast/added it by mistake

dbolson (approved):

PR #276, opened 2024-05-13
fix: add back member invite alias

k3llymariee:
This was accidentally removed when we generated the remaining resources from the openapi spec

Running `ldcli members --help` shows the new command:
```
Usage:
  ldcli members [command]

Available Commands:
  create       Invite new members
  create-teams Add a member to teams
  delete       Delete account member
  get          Get account member
  invite       Invite new members
  list         List account members
  update       Modify an account member

Global flags:
  -h, --help                  Get help about any command
      --access-token string   LaunchDarkly access token with write-level access
      --analytics-opt-out     Opt out of analytics tracking
      --base-uri string       LaunchDarkly base URI (default "https://app.launchdarkly.com")
  -o, --output string         Command response output format in either JSON or plain text (default "plaintext")

Use "ldcli members [command] --help" for more information about a command.
```

Note that the short description is the same as what we have for `create`, which comes straight from the openapi spec. Not sure what else we might want to put here instead.

### DIFF
--- a/cmd/members/invite.go
+++ b/cmd/members/invite.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	
+
 	"ldcli/cmd/cliflags"
 	resourcescmd "ldcli/cmd/resources"
 	"ldcli/cmd/validators"
--- a/cmd/members/invite_test.go
+++ b/cmd/members/invite_test.go
@@ -1,0 +1,49 @@
+package members_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"ldcli/cmd"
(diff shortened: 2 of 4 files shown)


cmd/members/invite.go
```diff
@@ -0,0 +1,88 @@
...
+	"ldcli/internal/output"
+	"ldcli/internal/resources"
+)
+
+func NewMembersInviteCmd(client resources.Client) *cobra.Command {
+	cmd := &cobra.Command{
+		Args:  validators.Validate(),
+		Long:  "Create new members and send them an invitation email",
+		RunE:  runE(client),
+		Short: "Invite new members",
```
k3llymariee, line 24: see comment in pr description: this is the same as what we have for `create`, which comes straight from the openapi spec. Not sure what else we might want to put here instead.

dbolson (approved):

PR #284, opened 2024-05-15
fix: remove deprecated user cmds

k3llymariee:


cmd/resources/resources.go
```diff
@@ -188,6 +188,7 @@ func NewResources(tags openapi3.Tags) map[string]ResourceData {
 
 func shouldFilter(name string) bool {
 	return strings.Contains(name, "(beta)") ||
+		strings.Contains(name, "User") ||
```
k3llymariee, line 191: includes `Users`, `User flag settings`, and `Users (beta)` for if/when we add in the beta endpoints

dbolson (approved):

PR #288, opened 2024-05-15
ci: enforce formatting, go mod tidy

cwaldren-ld:
Adds CI checks for:
- `gofmt`: important to ensure code is always formatted in the repo
- `go mod tidy`: important to ensure deterministic dependencies

It does this via `git/hooks/pre-commit`, which can also be installed via `make install-hooks`.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -21,22 +21,9 @@ jobs:
       with:
         go-version: '1.20'
 
-    - name: Enforce gofmt
+    - name: Enforce formatting / go mod tidy
       run: |
-        test -z $(gofmt -l cmd)
-        test -z $(gofmt -l internal)
-        test -z $(gofmt -l *.go)
-
-    - name: Enforce go mod tidy
-      run: |
-        set -e
-        go mod tidy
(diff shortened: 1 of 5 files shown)


.github/workflows/go.yml
```diff
@@ -21,6 +21,23 @@ jobs:
       with:
         go-version: '1.20'
 
+    - name: Enforce gofmt
+      run: |
+        test -z $(gofmt -l cmd)
```
cwaldren-ld, line 26: This is harder than it needs to be because of `vendor`, which has a lot of violations  - could also reverse it and test everything but filter out vendor results.

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.18.2
 	github.com/stretchr/testify v1.9.0
```
cwaldren-ld, line 20: (from `go mod tidy`)

cwaldren-ld:
Ok, refactored to be a pre-commit hook.

dbolson (approved):

PR #307, opened 2024-05-28
feat: add beta endpoints tor resource commands

k3llymariee:
Adds the `(beta)` tagged endpoints to our resources commands. Beta resources will all have `-beta` appended to their resource name. 

```
Available commands:
  access-tokens
  account-usage-beta
  applications-beta
  approval-requests
  audit-log
  code-refs
  context-settings
  contexts
  custom-roles
  data-export-destinations
  environments
  experiments-beta
  flag-links-beta
  flag-triggers
  flags
  flags-beta
  follow-flags
  integration-delivery-configurations-beta
  integration-subscriptions
  integrations-beta
  members
  members-beta
  metric-groups-beta
  metrics
  projects
  relay-proxy-configs
  release-pipelines-beta
  releases-beta
  scheduled-changes
  segments
  segments-beta
  tags
  teams
  teams-beta
  webhooks
  workflow-templates
  workflows

Use "ldcli [command] --help" for more information about a command.
```


cmd/resources/resource_utils.go
```diff
@@ -175,3 +189,7 @@ func getFlagName(paramName string) string {
 	}
 	return flagName
 }
+
+func replaceBackticks(s string) string {
+	return strings.Replace(s, "`", "'", -1)
+}
```
k3llymariee, line 195: for some weird reason cobra [takes the param description and adds it as the type](https://github.com/spf13/pflag/blob/d5e0c0615acee7028e1e2740a11102313be88de1/flag.go#L573):
```
Params: []Param{
	{
		Name:        "type",
		In:          "path",
		Description: "The type of event to retrieve. Must be either `received` or `published`.",
		Type:        "string",
	},
}
```
Screenshot 2024-05-28 at 1 20 09 PM

so as a hack I'm just replacing the backticks with `'`'s. This does update some existing commands but it shouldn't be a breaking change.

cmd/resources/resources.go
```diff
@@ -119,32 +120,43 @@ func GetTemplateData(fileName string) (TemplateData, error) {
 				requiresBody = op.RequestBody.Value.Required
 			}
 
+			var isBeta bool
+			if strings.Contains(tag, "(beta)") {
```
dbolson, line 124: This could be shortened `isBeta := strings.Contains(tag, "(beta)")`

internal/resources/client.go
```diff
@@ -24,13 +24,16 @@ func NewClient(cliVersion string) ResourcesClient {
 	return ResourcesClient{cliVersion: cliVersion}
 }
 
-func (c ResourcesClient) MakeRequest(accessToken, method, path, contentType string, query url.Values, data []byte) ([]byte, error) {
+func (c ResourcesClient) MakeRequest(accessToken, method, path, contentType string, query url.Values, data []byte, isBeta bool) ([]byte, error) {
```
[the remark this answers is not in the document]
dbolson: We could refactor to something else later if we find it's hard to work with.

dbolson (approved):

PR #321, opened 2024-06-17
chore: sc-240965/unauthed client

dbolson:
Creates a client for the login command that doesn't need an access token because the endpoint doesn't require it.

We then show the verification URL for the user to go to. Additional PRs will let the user open their browser with the URL and poll for a response.

**Requirements**

**Related issues**

**Additional context**


cmd/login/login.go
```diff
@@ -46,8 +53,31 @@ func NewLoginCmd(
...
+func run(client login.Client) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
+		deviceAuthorization, err := login.FetchDeviceAuthorization(
+			client,
+			login.ClientID,
+			login.GetDeviceName(),
+			viper.GetString(cliflags.BaseURIFlag),
+		)
+		if err != nil {
+			return errors.NewError(output.CmdOutputError(viper.GetString(cliflags.OutputFlag), err))
```
moribellamy, line 65: looks like `errors.NewError(output.CmdOutputError(viper.GetString(cliflags.OutputFlag), err))` appears 11 times in the project -- might be a good candidate for a function

internal/login/login.go
```diff
@@ -0,0 +1,106 @@
...
+	cliVersion string
+}
+
+func NewClient(cliVersion string) Client {
+	return Client{
+		cliVersion: cliVersion,
+	}
+}
+
+func (c Client) MakeRequest(
```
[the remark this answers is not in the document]
moribellamy: SGTM. if its not appropriate to do now but might be later, a comment might be helpful

internal/login/login.go
```diff
@@ -0,0 +1,106 @@
...
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
```
moribellamy, line 61: this should be called as soon as we know `res.Body` is not nil in case there is an error with reading, right?

moribellamy (approved):
lgtm with comments, feel free to wait for @k3llymariee

internal/login/login.go
```diff
@@ -0,0 +1,106 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+
+	"github.com/launchdarkly/ldcli/internal/errors"
+)
+
+const ClientID = "e6506150369268abae3ed46152687201"
```
k3llymariee, line 14: do we want to commit this clientID?

k3llymariee (approved):

internal/login/login.go
dbolson: I believe the idea is that res.Body will be nil if there's an error, so there's nothing to close. There's a [linter](https://github.com/timakin/bodyclose) with an example that closes the body after the error check.

internal/login/login.go
dbolson: I believe it should be safe. The risk would be if we need to rotate it, which I don't think we would since it's more for identity than security -- another client (like vscode) would have its own client ID. If someone were to make a request with this value, they wouldn't get any valuable data in the response.

cmd/login/login.go
dbolson: Yep, sounds good.

PR #327, opened 2024-06-24
(WIP): add command to start server

k3llymariee:
Adds a new top-level command for the `dev-server` (final name can be changed easily) and a subcommand to `start` the server.

Passing in our existing resources client to make any API request - and just showing that we will have access to whatever access token value is passed in from the config. 

start_server


internal/dev_server/dev_server.go
```diff
@@ -0,0 +1,9 @@
+package dev_server
+
+import (
+	"log"
+)
+
+func HelloWorld(accessTokenStr string) {
```
mike-zorn, line 7: I think a good next step here would be to port what's in dev_server/main.go into here.

mike-zorn: I'm also realizing it's probably better to pass around an api client instead of the api key. That way it'll work with different base urls.

k3llymariee: oh yeah we can/will, was just passing this along now to prove that we have it. the config file also already stores your base URL as well. 

here's an example of a request with the existing client we have: https://github.com/launchdarkly/ldcli/blob/f24a1ead0d5cd78c90d2b8e5bd754ca17678427a/cmd/flags/archive.go#L35-L49

internal/dev_server/dev_server.go
```diff
@@ -10,8 +10,7 @@ import (
 	"github.com/launchdarkly/ldcli/internal/dev_server/db"
 )
 
-// FIXME this is gonna get replaced by a cli command
-func main() {
+func RunServer() {
```
k3llymariee, line 13: @mike-zorn rather than porting this over I think it maybe makes more sense to leave this here and export it into the cmd, what do you think?

internal/dev_server/dev_server.go
k3llymariee: (but also I don't think we need a client for this cmd)

internal/dev_server/dev_server.go
mike-zorn: works for me!

k3llymariee:
merged into our moonshot-ld-dev-server branch

PR #328, opened 2024-06-25
Adding get project by project key

ld-kyee:
I'm unclear why there were so many formatting changes in the api.yaml file.

### DIFF
--- a/internal/dev_server/api/server.gen.go
+++ b/internal/dev_server/api/server.gen.go
@@ -448,6 +448,14 @@ func (response GetDevProjectsProjectKey200JSONResponse) VisitGetDevProjectsProje
 	return json.NewEncoder(w).Encode(response)
 }
 
+type GetDevProjectsProjectKey404Response struct {
+}
+
+func (response GetDevProjectsProjectKey404Response) VisitGetDevProjectsProjectKeyResponse(w http.ResponseWriter) error {
+	w.WriteHeader(404)
+	return nil
+}
+
 type PostDevProjectsProjectKeyRequestObject struct {
 	ProjectKey ProjectKey `json:"projectKey"`
 	Body       *PostDevProjectsProjectKeyJSONRequestBody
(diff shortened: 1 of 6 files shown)


internal/dev_server/api/server.gen.go
```diff
@@ -1,6 +1,6 @@
 // Package api provides primitives to interact with the openapi HTTP API.
 //
-// Code generated by github.com/oapi-codegen/oapi-codegen/v2 version v2.3.0 DO NOT EDIT.
+// Code generated by github.com/deepmap/oapi-codegen version v1.16.3 DO NOT EDIT.
```
mike-zorn, line 3: hmmmm how did you install the code generator? It looks like somehow you're using an old version.

internal/dev_server/dev_server.go
```diff
@@ -41,6 +42,7 @@ func (c LDClient) RunServer(ctx context.Context, accessToken, baseURI string) {
 	r.Use(model.StoreMiddleware(sqlStore))
 	// TODO need a subrouter for relay endpoints
 	handler := api.HandlerFromMux(apiServer, r)
+	fmt.Println("Server running on 0.0.0.0:8765")
```
mike-zorn, line 45: Good thinking!

mike-zorn (commented):
api yaml format changes are fine. Let's get the tool version figured out before you merge though.

internal/dev_server/api/server.gen.go
ld-kyee: 🤦 I thought I just did a `go install` but I had so many weird things going on locally.  Let me fix this.

mike-zorn: Oh the docs are kinda not great because they first tell you how to install the deprecated version. Try `go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest`

mike-zorn (approved):
approving, assuming we get the tool version straight.

PR #331, opened 2024-06-25
add override

k3llymariee:
image


internal/dev_server/api/server.go
```diff
@@ -50,11 +50,18 @@ func (s Server) PostDevProjectsProjectKey(ctx context.Context, request PostDevPr
 }
 
 func (s Server) DeleteDevProjectsProjectKeyOverridesFlagKey(ctx context.Context, request DeleteDevProjectsProjectKeyOverridesFlagKeyRequestObject) (DeleteDevProjectsProjectKeyOverridesFlagKeyResponseObject, error) {
-	//TODO implement me
-	panic("implement me")
+	store := model.StoreFromContext(ctx)
+	err := store.DeleteOverride(ctx, request.ProjectKey, request.FlagKey)
+	return DeleteDevProjectsProjectKeyOverridesFlagKey200JSONResponse{}, err
```
k3llymariee, line 55: not filling out this response body - it'll just default to false/null anyway which feels ok?

cmd/dev_server/projects.go
```diff
@@ -162,13 +162,13 @@ func NewAddProjectCmd(client dev_server.LocalClient) *cobra.Command {
 //}
 
 type postBody struct {
-	sourceEnvironmentKey string
+	SourceEnvironmentKey string `json:"sourceEnvironmentKey"`
```
k3llymariee, line 165: smol fix

internal/dev_server/api/server.go
ld-kyee: Maybe a 204 would be better if you're expecting an empty response?

.gitignore
```diff
@@ -2,3 +2,4 @@
 dist/
 *.log
 node_modules/
+devserver.db
```
mike-zorn, line 5: 🙌🏻

internal/dev_server/api/server.go
mike-zorn: We should either 204 or keep it 200 & return like `{Override: false, Value:...}`.

internal/dev_server/db/sqlite.go
```diff
@@ -49,6 +49,39 @@ VALUES (?, ?, ?, ?, ?)
...
+		valueJson,
+		override.Active,
+	)
+
+	return err
+}
+
+func (s Sqlite) DeleteOverride(ctx context.Context, projectKey, flagKey string) error {
+	_, err := s.database.Exec(`
+		UPDATE overrides set active = false where project_key = ? and flag_key = ?
```
mike-zorn, line 77: This should also increment the version. This is needed so that the sdk will pick up the change (it ignores changes if the version number isn't bigger).

k3llymariee: oh duh my b!

ld-kyee (approved):
When upserting we probably want to check the flag exists in the flagState for the project if not there will just be bogus entries if someone mistypes a flag name.  

Also a bonus would be to make sure the types of the flag values match but I know that's kinda a lot of work so maybe this can be done if we have time.

I'm not clear on how we are going to override values for boolean flags.  I tried setting the body to `{ "enable-deployment-link": "false"}` but it stored this entire thing in the value column which I don't think we want.  That said, I'm sure this is just me not understanding what the api expects.

This seems fine enough to merge for now and build off of if you'd rather add the validation check in a separate pr.

internal/dev_server/api/server.go
k3llymariee: updated to 204

PR #344, opened 2024-06-27
feat: Move devserver db to XDG_STATE_HOME

mike-zorn:
Use `XDG_STATE_HOME` for db location.

### DIFF
--- a/internal/dev_server/dev_server.go
+++ b/internal/dev_server/dev_server.go
@@ -6,10 +6,12 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"path/filepath"
 
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"github.com/launchdarkly/ldcli/internal/dev_server/sdk"
+	"github.com/mitchellh/go-homedir"
 
 	"github.com/launchdarkly/ldcli/internal/client"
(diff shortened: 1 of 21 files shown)


hoshsadiq (approved):

internal/dev_server/dev_server.go
```diff
@@ -67,3 +71,15 @@ func RequestErrorHandler(w http.ResponseWriter, r *http.Request, err error) {
 	log.Printf("Error while serving request: %+v", err)
 	http.Error(w, err.Error(), http.StatusBadRequest)
 }
+
+func getDBPath() string {
+	statePath := os.Getenv("XDG_STATE_HOME")
+	if statePath == "" {
+		home, _ := homedir.Dir()
+		statePath = filepath.Join(home, ".local/state")
+	}
```
hoshsadiq, line 80: By the way, just realised. This translates to `~/.local/state/dev_server.db`, we probably want `~/.local/state/ldcli/dev_server.db` or something.

Worth using [`xdg.StateFile("ldcli/dev_server.db")`](https://github.com/adrg/xdg) instead.

mike-zorn: Oh! Nice. Yeah. I'll do this instead.

PR #346, opened 2024-06-27
Adds a sync api

ld-kyee:
I've added a sync api for the project that refetches the flag state.

The api looks like this 

`PATCH localhost:8765/dev/projects/default/sync?expand=overrides`

I've updated it so that the expand param is in both the patch and post apis since we'll want that data for the ui.  

In the next pr, I'll add another endpoint to patch the context and source environment for the project.  I should also do a bit of refactoring.  😬

### DIFF
--- a/internal/dev_server/model/project.go
+++ b/internal/dev_server/model/project.go
@@ -64,3 +64,31 @@ func (p Project) GetFlagStateWithOverridesForProject(ctx context.Context) (Flags
 	}
 	return withOverrides, nil
 }
+
+func SyncProject(ctx context.Context, projectKey string) (Project, error) {
+	store := StoreFromContext(ctx)
+	project, err := store.GetDevProject(ctx, projectKey)
+	if err != nil {
+		return Project{}, err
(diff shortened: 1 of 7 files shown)


internal/dev_server/model/project.go
```diff
@@ -48,6 +48,34 @@ func CreateProject(ctx context.Context, projectKey, sourceEnvironmentKey string,
...
+	sdkKey, err := apiAdapter.GetSdkKey(ctx, projectKey, project.SourceEnvironmentKey)
+	if err != nil {
+		return Project{}, err
+	}
+	sdkFlags, err := sdkAdapter.GetAllFlagsState(ctx, project.Context, sdkKey)
+	if err != nil {
+		return Project{}, err
+	}
+	project.FlagState = FromAllFlags(sdkFlags)
+	project.LastSyncTime = time.Now()
```
mike-zorn, line 68: Maybe we can extract this into a method on `*Project` that can be reused by the `CreateProject`

PR #348, opened 2024-06-27
feat: add ui routes

mike-zorn:
Adds UI routes & uses go generate instead of make for regenerating the server code.

### DIFF
--- a/internal/dev_server/ui/asset_handler.go
+++ b/internal/dev_server/ui/asset_handler.go
@@ -1,0 +1,11 @@
+package ui
+
+import (
+	"embed"
+	"net/http"
+)
+
+//go:embed index.html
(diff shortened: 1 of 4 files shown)


internal/dev_server/ui/asset_handler.go
```diff
@@ -0,0 +1,11 @@
+package ui
+
+import (
+	"embed"
+	"net/http"
+)
+
+//go:embed index.html
```
mike-zorn, line 8: @kwatkins-ld you can add more assets to the embed if you list them here. Take a look at these [docs](https://pkg.go.dev/embed#hdr-Directives) for the syntax.

kwatkins-ld (approved):

PR #352, opened 2024-06-27
Add project patch

ld-kyee:
This allows contexts and sourceEnvironmentKey to be patched on the project and I also updated the post to allow context creation.  I did not update the ldcli yet.

@mike-zorn I wasn't a 100% sure how to check if the contexts were correctly being applied.  I'm not too clear on how marshalling works and if things needed to be marshalled or unmarshalled or what not.  So if you can take a look at that, that would be great.

### DIFF
--- a/internal/dev_server/api/api.yaml
+++ b/internal/dev_server/api/api.yaml
@@ -28,15 +28,7 @@ paths:
       summary: updates the flag state for the given project and source environment
       parameters:
         - $ref: "#/components/parameters/projectKey"
-        - name: expand
-          description: Available expand options for this endpoint.
-          in: query
-          schema:
-            type: array
-            items:
-              type: string
-              enum:
-                - overrides
+        - $ref: "#/components/parameters/expand"
       responses:
         200:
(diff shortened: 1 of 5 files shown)


internal/dev_server/api/api.yaml
```diff
@@ -61,7 +61,35 @@ paths:
...
+        - $ref: "#/components/parameters/projectKey"
+        - name: expand
+          description: Available expand options for this endpoint.
+          in: query
+          schema:
+            type: array
+            items:
+              type: string
+              enum:
+                - overrides
```
mike-zorn, line 76: Should we move this into components?

mike-zorn:
Marshaling looks good to me! I think you could do a manual test by copying [a context out of catfood](https://app.ld.catamorphic.com/projects/default/contexts/instances/dc07cbc3-4e44-5bcb-8424-5358a6cb7ed9?env=production&env=staging&selected-env=staging) and trying to post it.

mike-zorn (approved):
one note on the api spec reuse, but otherwise looks good!

PR #368, opened 2024-07-03
chore: Ensure valid base URI

dbolson:
If the user sets the base URI to have a trailing slash, we end up with a 301 and the actual request, which is extra work in the back end. Instead of building the URLs with `fmt.Sprintf()` we can use `url.JoinPath()` to remove the extra trailing slash. This is also simpler than ensuring the user's base URI value doesn't have a trailing slash.
**Requirements**

**Related issues**

**Additional context**


cmd/resources/resources.go
```diff
@@ -283,6 +283,7 @@ func (op *OperationCmd) initFlags() error {
 }
 
 func buildURLWithParams(baseURI, path string, urlParams []string) string {
```
dbolson, line 285: There aren't tests for this, and I didn't want to possibly break things, so this will remove the trailing slash or be a no-op.

k3llymariee (approved):

PR #369, opened 2024-07-03
use resource client instead of separate dev server client

k3llymariee:
Danny just added a new `MakeUnauthenticatedRequest` method on our existing client we use for other API requests, so we can use that instead of spinning up our own client. 

Also merges main into our branch.


mike-zorn (approved):
✨✨✨✨
✨LGTM✨
✨✨✨✨

PR #376, opened 2024-07-10
fix(deps): build with Go 1.22

cwaldren-ld:
Bumps the Go version in `go.mod` to 1.22. Additionally, installs Go 1.22 on the github runner (before, it must've been using whatever the runner had available by default.)


.github/actions/publish/action.yml
```diff
@@ -28,6 +28,9 @@ runs:
       uses: docker/setup-buildx-action@v3
       with:
         platforms: linux/amd64,linux/arm64/v8,linux/arm/v7,linux/386
```
cwaldren-ld, line 30: Since we weren't specifying this before, our release binaries would be using whatever version of Go is on the Github Runner

PR #377, opened 2024-07-11
fix(deps): use consistent Go version in builds and CI

cwaldren-ld:
Currently we are building with whatever Go version is installed on the runner at the time of the build. At the same time, we're testing with a particular Go version in CI.

This PR unifies the versions used in test and build. That way we know we've tested what we're shipping.

- [x] PR has a matrix build of go versions

### DIFF
--- a/.github/workflows/go-versions.yml
+++ b/.github/workflows/go-versions.yml
@@ -1,0 +1,49 @@
+# The following chunk of yml boils down to pulling two Go version numbers out of a file and
+# making them available to the other workflows in a convenient fashion.
+#
+# It's a reusable workflow instead of an action so that its output can be used in a matrix strategy
+# of another job.
+#
+# The idea is to define the most recent, and penultimate, Go versions that should be used to test Relay.
+# Ideally we'd define these in a single place - otherwise we'd need to update many different places in
+# each workflow. This single place is .github/variables/go-versions.env.
+#
+# This reusable workflow grabs them out of the file, then sets them as outputs. As a convenience, it
+# also wraps each version in an array, so it can be directly used in a matrix strategy. Single-item matrices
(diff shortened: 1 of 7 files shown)


.github/workflows/go-versions.yml
```diff
@@ -0,0 +1,49 @@
...
+      all: ${{ steps.set-matrix.outputs.all }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set Go Versions
+        id: set-env
+        run:  cat ./.github/variables/go-versions.env > $GITHUB_OUTPUT
+      - name: Set Go Version Matrices
+        id: set-matrix
+        run: |
+          echo "all=[\"${{ steps.set-env.outputs.latest }}\",\"${{ steps.set-env.outputs.penultimate }}\"]" >> $GITHUB_OUTPUT
```
cwaldren-ld, line 49: (This whole file is copied from Relay. It'd be nice to make it reusable.. but not sure it's worth it yet.)

cwaldren-ld:
There is a choice we can make for supporting Go versions. We can support the latest release N and penultimate N-1, and keep these up-to-date automatically. 

This would be equivalent to roughly 1 year of Go support. 

Or, we could support latest release N and some other minimum version < N, not necessarily the penultimate. This allows for wider compatibility, at the cost of having to think about when to bump the minimum version.

Since we're shipping binaries of this, and people likely aren't building from source, I think the first option is easiest.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/launchdarkly/ldcli
 
-go 1.20
+go 1.21
```
cwaldren-ld, line 3: Although it's not strictly true that we require 1.21, since we're building with 1.21 and 1.22 we can't claim this. See my other comment for an alternative.

dbolson (approved):

PR #378, opened 2024-07-11
feat: use SDK metadata

cwaldren-ld:
The CLI currently hard-codes certain information about LaunchDarkly SDKs, including example links, display names, and SDK types (server, client, and so on.)

This information must be manually kept up-to-date by the CLI maintainers or SDK engineers. 

With the introduction of the [`sdk-meta`](https://github.com/launchdarkly/sdk-meta) Go module, we're able to source this metadata externally. 

As background, whenever a LaunchDarkly SDK's metadata is updated ([example](https://github.com/launchdarkly/js-core/blob/main/.sdk_metadata.json)), the sdk-meta project [crawls the changes](https://github.com/launchdarkly/sdk-meta/pull/129) and generates a [new Go module version](https://github.com/launchdarkly/sdk-meta/pull/130). 

------------------

In this PR, I've removed data that is now provided by the sdk-meta module. SDK example links are not yet part of the sdk-meta project, so I transformed those into a simple lookup table. 

Finally, I renamed the SDK quickstart instruction files. To retrieve metadata for a particular SDK, you need the SDK's ID. It makes sense to define this as part of the filename. 

The effect is that if you add a new quickstart instruction with a particular SDK ID, metadata will be automatically found for it.


cwaldren-ld:
The majority of the changes are `vendor` updates.

internal/quickstart/choose_sdk.go
```diff
@@ -61,15 +62,84 @@ func NewChooseSDKModel(selectedIndex int) tea.Model {
...
+	"react-client-sdk":  "https://github.com/launchdarkly/react-client-sdk/tree/main/examples/typescript",
+	"vue":               "https://github.com/launchdarkly/vue-client-sdk/tree/main/example",
+	"react-native":      "https://github.com/launchdarkly/js-core/tree/main/packages/sdk/react-native/example",
+	"cpp-client-sdk":    "https://github.com/launchdarkly/cpp-sdks/tree/main/examples/hello-cpp-client",
+	"cpp-server-sdk":    "https://github.com/launchdarkly/cpp-sdks/tree/main/examples/hello-cpp-server",
+	"lua-server-server": "https://github.com/launchdarkly/lua-server-sdk/tree/main/examples/hello-lua-server",
+}
+
+// sdkOrder is a list of IDs the SDKs in order that they should be rendered based on popularity.
+var sdkOrder = []string{
```
dbolson, line 84: I added back the order in which the SDKs should show up based on popularity. Otherwise they show up in alphabetical order.

internal/quickstart/choose_sdk.go
```diff
@@ -18,23 +24,18 @@ var (
...
 	list          list.Model
+	sdkDetails    []sdkDetail
 	selectedIndex int
 	selectedSDK   sdkDetail
 }
 
 func NewChooseSDKModel(selectedIndex int) tea.Model {
-	initSDKs()
-	l := list.New(sdksToItems(), sdkDelegate{}, 30, 9)
+	sdkDetails := initSDKs()
```
dbolson, line 37: I updated the PR to not use the global var but pass this list around instead. It felt weird to modify it when we could encapsulate it in the model.

internal/quickstart/choose_sdk.go
```diff
@@ -61,15 +62,90 @@ func NewChooseSDKModel(selectedIndex int) tea.Model {
...
+
+	details := make([]sdkDetail, 0, len(items))
+	for _, item := range items {
+		id, _, _ := strings.Cut(filepath.Base(item.Name()), ".")
+		if _, ok := sdkmeta.Names[id]; !ok {
+			continue
+		}
+		index := slices.Index(sdkOrder, id)
+		if index == -1 {
+			// if we missed an SDK don't add it with an invalid index
```
dbolson, line 135: Without this, there's a panic since it would later try to reference a -1 index. It _shouldn't_ happen unless we change the SDK ID and forget to update the order list.

dbolson (approved):

internal/quickstart/choose_sdk.go
cwaldren-ld: Ah, that's interesting - I guess sdk-meta needs to maintain a popularity list as well. I'll make a ticket.

PR #388, opened 2024-08-13
fix: Get build working (dev server)

mike-zorn:

Removes the normal sqlite library in favor of one that doesn't use CGO.

Tried using different methods of multi-platform builds
- musl & clang & mingw
- zig -- ran into an issue where the macos builds wouldn't work as described
  [here](https://forum.golangbridge.org/t/cgo-zig-netgo-and-libresolv/33842).

**Additional context**

Note that this includes an extraneous workflow that I was using to test the
build. I'll remove prior to merging.


mike-zorn:
Closing in favor of https://github.com/launchdarkly/ldcli/pull/389

PR #403, opened 2024-08-16
fix: rebuild UI to incorporate sync button

mike-zorn:
The prior PR to add the sync button didn't have any effect because the UI
wasn't built.


cdelst (approved):
Darn

PR #412, opened 2024-08-24
fix: Propagate dev server errors in CLI compatible format

mike-zorn:
This makes it so that errors from the dev server are propagated back to the CLI in a way that the CLI can reasonably render. This way, we can return something that gives the user an idea of what went wrong vs. a generic "unexpected error" message.

example
```
❯ ldcli dev-server add-project --project jimbo-jones --source staging
unable to get SDK key from LD API: 404 Not Found (code: internal_server_error)
```

### DIFF
--- a/internal/dev_server/api/error_handlers.go
+++ b/internal/dev_server/api/error_handlers.go
@@ -25,11 +25,15 @@ func (eh errorHandler) HandleError(w http.ResponseWriter, r *http.Request, err e
 }
 
 var RequestErrorHandler = errorHandler{
+	// HACK: This is really just repeating the status code.
+	// It'd be nice to make these be codes that are meaningful to the user.
 	code:       "bad_request",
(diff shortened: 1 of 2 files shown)


internal/dev_server/api/error_handlers.go
```diff
@@ -0,0 +1,35 @@
+package api
+
+import (
+	"encoding/json"
+	"log"
+	"net/http"
+)
+
+type errorHandler struct {
+	code       string
```
kinyoklion, line 10: nit: Code seems like a strange name for this. It *looks* like statusText.

kinyoklion (approved):

dbolson (approved):

internal/dev_server/api/error_handlers.go
kinyoklion: Cool. I figured there was a reason for it.

PR #418, opened 2024-09-04
feat: Display source environment in LaunchDevly UI

cdelst:
**Requirements**

image

### DIFF
--- a/internal/dev_server/ui/src/Flags.tsx
+++ b/internal/dev_server/ui/src/Flags.tsx
@@ -2,7 +2,7 @@ import { LDFlagSet, LDFlagValue } from 'launchdarkly-js-client-sdk';
 import { Button, Checkbox, IconButton, Label } from '@launchpad-ui/components';
 import { Box, CopyToClipboard } from '@launchpad-ui/core';
 import Theme from '@launchpad-ui/tokens';
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useCallback, useMemo } from 'react';
 import { Icon } from '@launchpad-ui/icons';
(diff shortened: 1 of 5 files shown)


internal/dev_server/ui/src/Flags.tsx
```diff
@@ -23,61 +29,70 @@ function Flags({ selectedProject, flags, setFlags }: FlagProps) {
     Record<string, FlagVariation[]>
   >({});
 
-  const overridesPresent = overrides && Object.keys(overrides).length > 0;
+  const overridesPresent = useMemo(
```
cdelst, line 32: best practice to not recalculate these every time

internal/dev_server/ui/src/Flags.tsx
```diff
@@ -23,61 +29,70 @@ function Flags({ selectedProject, flags, setFlags }: FlagProps) {
...
+    [overrides],
+  );
 
-  const updateOverride = (flagKey: string, overrideValue: LDFlagValue) => {
-    const updatedOverrides = {
-      ...overrides,
-      ...{
-        [flagKey]: {
-          value: overrideValue,
+  const updateOverride = useCallback(
```
cdelst, line 37: useCallback lets us cache this func definition between renders

mike-zorn: Do we want that though? We closure in the override state...

mike-zorn: ohhh, I see you also pass in the deps.

mike-zorn (approved):

PR #420, opened 2024-09-06
feat: Pick from list of real variations

mike-zorn:
This adds better UX for multivariate flags. Previously, you just had the ability to edit the raw json values. Now you can pick from the available variations already defined in the LD service or you can create your own "local override". This is implemented by fetching all flags from the flag list API and writing the variations to the sqlite db. The data written to SQLite is then used to serve subsequent requests. I took this approach because it means that the pulled flag state should be in-sync with the available variations and it enables offline work. It does mean that add project & sync project operations take significantly longer in projects with large amounts of flags.

Some other changes are along for the ride because this branch was longer lived than ideal.

- search bar for flags
- display source environment
- dark mode
- fix patch endpoint so that the side-effect resync causes the observers to fire


internal/dev_server/db/sqlite.go
```diff
@@ -70,7 +72,16 @@ func (s Sqlite) UpdateProject(ctx context.Context, project model.Project) (bool,
 		return false, errors.Wrap(err, "unable to marshal flags state when updating project")
 	}
 
-	result, err := s.database.ExecContext(ctx, `
+	tx, err := s.database.BeginTx(ctx, nil)
```
cdelst, line 75: sqllite has transactions???

cdelst (approved):

internal/dev_server/db/sqlite.go
mike-zorn: Yeah. SQLite is kinda incredible. It's one of the best [tested](https://www.sqlite.org/testing.html) and [widely used](https://www.sqlite.org/mostdeployed.html) pieces of software you'll find.

PR #426, opened 2024-09-12
fix: LaunchDevly: Pagination for 'Only show flags with overrides'

cdelst:
Fixes the interaction with pagination when toggling the "only" toggle.  Puts the filter ahead of the display logic so that our pagination doesn't get confused why the render is leaving flags out.  

This works as expected now.

### DIFF
--- a/internal/dev_server/ui/src/Flags.tsx
+++ b/internal/dev_server/ui/src/Flags.tsx
@@ -53,12 +53,24 @@ function Flags({
   const filteredFlags = useMemo(() => {
     if (!flags) return [];
     const flagEntries = Object.entries(flags);
-    const search = searchTerm.toLowerCase();
-    const filtered = fuzzysort
-      .go(search, flagEntries, { all: true, key: '0', threshold: 0.7 })
-      .map((result) => result.obj);
-    return filtered;
-  }, [flags, searchTerm]);
+    return flagEntries
+      .filter((entry) => {
+        if (!searchTerm) return true;
+        const [flagKey] = entry;
+        const result = fuzzysort.single(searchTerm.toLowerCase(), flagKey);
+        return result && result.score > -5000; // Adjust threshold as needed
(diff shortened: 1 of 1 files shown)


internal/dev_server/ui/src/Flags.tsx
```diff
@@ -231,6 +243,7 @@ function Flags({
                   setSearchTerm(e.target.value);
                   setCurrentPage(0); // Reset pagination
                 }}
+                value={searchTerm}
```
cdelst, line 246: Previously the input of the search box wasn't actually coupled to the display of the search... We were just relying on onChange events.

internal/dev_server/ui/src/Flags.tsx
```diff
@@ -247,10 +260,6 @@ function Flags({
             const hasOverride = flagKey in overrides;
             const currentValue = hasOverride ? overrideValue : flagValue;
 
-            if (onlyShowOverrides && !hasOverride) {
```
cdelst, line 250: This was causing issues, because we need to not do any filter logic in the display logic for the pagination to work as expected.

internal/dev_server/ui/src/Flags.tsx
```diff
@@ -53,12 +53,24 @@ function Flags({
...
     if (!flags) return [];
     const flagEntries = Object.entries(flags);
-    const search = searchTerm.toLowerCase();
-    const filtered = fuzzysort
-      .go(search, flagEntries, { all: true, key: '0', threshold: 0.7 })
-      .map((result) => result.obj);
-    return filtered;
-  }, [flags, searchTerm]);
+    return flagEntries
+      .filter((entry) => {
```
cdelst, line 57: Refactored to be clearer filter entries.  I assume we'll add another for tags in the future ;)

mike-zorn (approved):

PR #427, opened 2024-09-12
fix: LaunchDevly: Cleanup code and remove only show overrides when there are no overrides

cdelst:
Previously you could get stuck in an error state where the "only show overrides" checkbox would get locked to on, and then disabled so that you couldn't edit it again due to having no overrides.  

I solved this by adding a cheap useEffect hook that checks for this condition and directs us into a valid state. 

I thought about hooking this in in various places to avoid the useEffect usage, but a global watcher for this component made the most sense to me.

### DIFF
--- a/internal/dev_server/ui/src/Flags.tsx
+++ b/internal/dev_server/ui/src/Flags.tsx
@@ -16,7 +16,7 @@ import {
   Stack,
 } from '@launchpad-ui/core';
 import Theme from '@launchpad-ui/tokens';
-import { useState, useCallback, useMemo } from 'react';
+import { useState, useCallback, useMemo, useEffect } from 'react';
 import { Icon } from '@launchpad-ui/icons';
 import { apiRoute } from './util.ts';
 import { FlagVariation } from './api.ts';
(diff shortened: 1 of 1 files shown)


internal/dev_server/ui/src/Flags.tsx
```diff
@@ -207,24 +209,11 @@ function Flags({
...
-                // Opt out of local state updates since we're bulk-removing
-                // overrides async
-                removeOverride(flagKey);
+                return removeOverride(flagKey);
               }),
             );
 
-            // Winnow out removed overrides and update local state in a
-            // single pass
-            const updatedOverrides = overrideKeys.reduce(
```
cdelst, line 218: Turns out this was entirely unnecessary

internal/dev_server/ui/src/Flags.tsx
```diff
@@ -42,14 +42,20 @@ function Flags({
...
-  const [currentPage, setCurrentPage] = useState(0); // Change initial page to 0
+  const [currentPage, setCurrentPage] = useState(0);
   const flagsPerPage = 20;
 
   const overridesPresent = useMemo(
     () => overrides && Object.keys(overrides).length > 0,
     [overrides],
   );
 
+  useEffect(() => {
```
cdelst, line 53: Main bug fix here.

mike-zorn (approved):

PR #437, opened 2024-09-27
fix: Context updating, environment rate limiting, toggle styling

cdelst:
We now query environments by name, instead of all X at once up front.  

image

image

image

image

### DIFF
--- a/internal/dev_server/adapters/api.go
+++ b/internal/dev_server/adapters/api.go
@@ -2,6 +2,7 @@ package adapters
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"net/url"
 	"strconv"
@@ -25,7 +26,7 @@ func GetApi(ctx context.Context) Api {
 type Api interface {
 	GetSdkKey(ctx context.Context, projectKey, environmentKey string) (string, error)
 	GetAllFlags(ctx context.Context, projectKey string) ([]ldapi.FeatureFlag, error)
-	GetProjectEnvironments(ctx context.Context, projectKey string) ([]ldapi.Environment, error)
+	GetProjectEnvironments(ctx context.Context, projectKey string, query string) ([]ldapi.Environment, error)
 }
 
 type apiClientApi struct {
@@ -54,9 +55,9 @@ func (a apiClientApi) GetAllFlags(ctx context.Context, projectKey string) ([]lda
 	return flags, err
 }
 
-func (a apiClientApi) GetProjectEnvironments(ctx context.Context, projectKey string) ([]ldapi.Environment, error) {
+func (a apiClientApi) GetProjectEnvironments(ctx context.Context, projectKey string, query string) ([]ldapi.Environment, error) {
 	log.Printf("Fetching all environments for project '%s'", projectKey)
-	environments, err := a.getEnvironments(ctx, projectKey, nil)
+	environments, err := a.getEnvironments(ctx, projectKey, nil, query)
 	if err != nil {
 		err = errors.Wrap(err, "unable to get environments from LD API")
 	}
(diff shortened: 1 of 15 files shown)


internal/dev_server/ui/src/Flag.tsx
```diff
@@ -45,12 +47,18 @@ const VariationValues = ({
   switch (typeof flagValue) {
     case 'boolean':
       return (
-        <Switch
-          isSelected={currentValue}
-          onChange={(newValue) => {
-            updateOverride(flagKey, newValue);
-          }}
-        />
+        <div className="animated-switch-container">
```
cdelst, line 50: The styles broke for our launchpad component.  I want to get this fix out before debugging that, so for now I created a new style that styles correctly.

We were having issues with the styling regarding rems and correctly spacing the switch anyway, so this improves overall looks too at the cost of custom css.

internal/dev_server/ui/src/Switch.css
```diff
@@ -0,0 +1,77 @@
+
+.animated-switch-container {
```
cdelst, line 2: Going to be honest, most of this was written with Cursor.  It looks great to me, and functions perfectly, so not questioning it.  

It can be removed when we figure out the Switch issue in launchpad.

internal/dev_server/adapters/api.go
```diff
@@ -81,21 +82,24 @@ func (a apiClientApi) getFlags(ctx context.Context, projectKey string, href *str
 	})
 }
 
-func (a apiClientApi) getEnvironments(ctx context.Context, projectKey string, href *string) ([]ldapi.Environment, error) {
-	return getPaginatedItems(ctx, projectKey, href, func(ctx context.Context, projectKey string, limit, offset *int64) (*ldapi.Environments, error) {
-		request := a.apiClient.EnvironmentsApi.GetEnvironmentsByProject(ctx, projectKey)
-		if limit != nil {
-			request = request.Limit(*limit)
-		}
```
vroske-ld, line 89: i think we still need this in some form? i don't see the [limit](https://github.com/launchdarkly/ldcli/blob/3dcc22f0280e4b996739476c2bf2282a1ab13bcd/internal/dev_server/ui/src/api.ts#L26) from the frontend being respected

internal/dev_server/ui/src/EnvironmentSelector.tsx
```diff
@@ -17,51 +24,59 @@ export function EnvironmentSelector({
...
+  }, [fetchEnvironmentsDebounced, searchQuery]);
+
+  useEffect(() => {
+    const filtered = environments?.filter(
       (env) =>
         env.name.toLowerCase().includes(searchQuery.toLowerCase()) ||
         env.key.toLowerCase().includes(searchQuery.toLowerCase()),
     );
-    setFilteredEnvironments(filtered);
+    setFilteredEnvironments(filtered || []);
```
vroske-ld, line 77: it's not clear to me if we still need to have a reference to `filteredEnvironments`. it seems like the intention was to limit the number of envs we get upfront, so the filtering should now be happening on the backend rather than the frontend? please correct me if i've misunderstood 😅

vroske-ld (commented):
i have a couple of questions! also, i found an error when switching to a couple of specific environments. it might be outside the scope of this specific PR but wanted to raise what i found:

when trying to switch to production i got: 
```
PATCH http://localhost:5173/api/dev/projects/default 500
{
    "code": "internal_server_error",
    "message": "unable to get source flags from LD SDK: LaunchDarkly client initialization failed"
}
```

i also experienced this when switching to catfood, but not other environments like my own local dev, back to staging, etc.

internal/dev_server/ui/src/EnvironmentSelector.tsx
cdelst: Ooh yea I think you're right! The backend is now doing the filtering :)

internal/dev_server/adapters/api.go
cdelst: oh huh, interesting.  I'll debug this.  Thanks for testing and calling it out!

cdelst:
@vroske-ld I am making a ticket to investigate your unrelated issue.

internal/dev_server/adapters/api.go
cdelst: yea, turns out I forgot to add it back 😅

internal/dev_server/ui/src/EnvironmentSelector.tsx
cdelst: done

vroske-ld (approved):

PR #442, opened 2024-10-01
fix: LaunchDevly UI: Add testing framework

cdelst:
1. Add testing framework and simple unit test as example.
2. Add pre-commit checks that run the tests and make sure the UI is built


internal/dev_server/ui/package.json
```diff
@@ -2,19 +2,20 @@
...
   "version": "0.0.0",
-  "type": "module",
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "prettier:write": "prettier . --write",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "test": "jest --config jest.config.mjs"
```
vroske-ld, line 11: why not vite, since you're already using it to build the project? 😁

vroske-ld (approved):

internal/dev_server/ui/package.json
cdelst: ah very true, just used to it.  Will change

cdelst: fixed

PR #446, opened 2024-10-07
fix: Retry when there are 429s on flag metadata fetches

mike-zorn:
This adds retry-on-429 support to the dev server when it syncs data from the flags API. It uses the `X-RateLimit-Reset` to determine when to retry the request. This also moves the api utilities into an internal package instead of inline with the api adapter.


cdelst (approved):
cool beans

PR #450, opened 2024-10-17
feat(dev-server): Add toast for successful sync and warning for failure

cdelst:
image


internal/dev_server/ui/src/main.tsx
```diff
@@ -28,6 +29,7 @@ const Root = () => {
     <React.StrictMode>
       <IconProvider>
         <App />
```
mike-zorn, line 31: Should it be at the top? That's where most of the latent interactions happen.

mike-zorn (approved):
One thought on placement, but 👍🏻

internal/dev_server/ui/src/main.tsx
cdelst: TBH, not sure.  I think just having it somewhere relatively global is standard.

Don't think it matters that much here?

cdelst: wym by at the top?

mike-zorn: oh like at the top of the page. Screenshot makes it look like maybe it's at the bottom of the page?

mike-zorn: Location in the markup seems fine.

cdelst: ohhh, yea... bottom of the page is standard for toasts and I don't think We can change that.  The <ToastProvider> doesn't actually control that either.  The position is absolute no matter where we render the provider

cdelst: When using it it's clear enough to give some confirmation 👍🏻

mike-zorn: 🆒

PR #467, opened 2024-12-24
fix: Adjust payload format for PHP endpoints

keelerm84:


internal/dev_server/sdk/routes.go
```diff
@@ -21,6 +21,9 @@ func BindRoutes(router *mux.Router) {
 
 	router.Handle("/all", GetProjectKeyFromAuthorizationHeader(http.HandlerFunc(StreamServerAllPayload)))
 
+	router.PathPrefix("/sdk/flags/{flagKey}").
```
keelerm84, line 24: The `GetServerFlags` method already has a branch of logic that handles the individual flag request if the `flagKey` variable is set. However, we never defined that route so the variable would in fact get set.

internal/dev_server/sdk/get_server_flags.go
```diff
@@ -22,7 +22,7 @@ func GetServerFlags(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, "flag not found", http.StatusNotFound)
 		}
 	} else {
-		body = ServerAllPayloadFromFlagsState(allFlags)
+		body = ServerFlagsFromFlagsState(allFlags)
```
keelerm84, line 25: For PHP, the `/sdk/flags` endpoint should return only the flags in an array, not the full flags + segments payload.

mike-zorn (approved):
Thanks for fixing this!

internal/dev_server/sdk/get_server_flags.go
mike-zorn: Oh wait a sec. Will this break non-php sdks? I think we need this to have a 3rd case where we do just flags and not all?

keelerm84: I don't believe this is a problem. This function only serves two routes -- `/sdk/flags/{flagKey}` and `/sdk/flags`. 

These are routes that should exclusively be used by PHP SDKs. Non-PHP SDKs don't make flag only requests such as these.

> I think we need this to have a 3rd case where we do just flags and not all?

Line 25 is serves all flags. Line 20 serves a single flag (assuming the flag key parameter is set).

The `/all` endpoint is serviced by the `StreamServerAllPayload` handler.

mike-zorn: > The /all endpoint is serviced by the StreamServerAllPayload handler.

Ahhh, OK. Cool. I forgot about that.

mike-zorn (approved):

PR #485, opened 2025-02-03
feat: Skip syncing on start if flag set

tvarney13:
**Requirements**

**Related issues**
https://launchdarkly.atlassian.net/browse/FUN-447

Added a `should-sync` flag to the server startup command that sets `model.InitialProjectSettings.Enable` which controls whether [this function ](https://github.com/launchdarkly/ldcli/blob/FUN-447-start-n-sync-dont-blow-away-existing-db/internal/dev_server/model/sync.go#L24) no-ops or not.

Using the no-op made sense for the persistent volumes use case, since all the project and env data will already exist locally, but open to being more surgical if so desired.

We could skip the `UpdateProject` call specifically in the `CreateOrSyncProject` function, which does the sync bit. Then we'd have to decide if we still want to do the overrides.

**Additional context**

### DIFF
--- a/cmd/config/testdata/help.golden
+++ b/cmd/config/testdata/help.golden
@@ -10,6 +10,7 @@ Supported settings:
 - `output`: Command response output format in either JSON or plain text
 - `port`: Port for the dev server to run on
 - `project`: Default project key
+- `sync-once`: Only sync new projects. Existing projects will not be resynced on startup
 
 Usage:
   ldcli config [flags]
--- a/cmd/dev_server/start_server.go
+++ b/cmd/dev_server/start_server.go
@@ -43,6 +43,9 @@ func NewStartServerCmd(client dev_server.Client) *cobra.Command {
 	cmd.Flags().String(OverrideFlag, "", `Stringified JSON representation of flag overrides ex. {"flagName": true, "stringFlagName": "test" }`)
 	_ = viper.BindPFlag(OverrideFlag, cmd.Flags().Lookup(OverrideFlag))
 
+	cmd.Flags().Bool(cliflags.SyncOnStartFlag, true, cliflags.SyncOnStartFlagDescription)
+	_ = viper.BindPFlag(cliflags.SyncOnStartFlag, cmd.Flags().Lookup(cliflags.SyncOnStartFlag))
+
 	return cmd
 }
 
@@ -55,7 +58,7 @@ func startServer(client dev_server.Client) func(*cobra.Command, []string) error 
 		if viper.IsSet(cliflags.ProjectFlag) && viper.IsSet(SourceEnvironmentFlag) {
 
 			initialSetting = model.InitialProjectSettings{
-				Enabled:    true,
+				Enabled:    viper.GetBool(cliflags.SyncOnStartFlag),
 				ProjectKey: viper.GetString(cliflags.ProjectFlag),
 				EnvKey:     viper.GetString(SourceEnvironmentFlag),
 			}
--- a/cmd/dev_server/start_server.go
+++ b/cmd/dev_server/start_server.go
@@ -43,8 +43,8 @@ func NewStartServerCmd(client dev_server.Client) *cobra.Command {
 	cmd.Flags().String(OverrideFlag, "", `Stringified JSON representation of flag overrides ex. {"flagName": true, "stringFlagName": "test" }`)
 	_ = viper.BindPFlag(OverrideFlag, cmd.Flags().Lookup(OverrideFlag))
 
-	cmd.Flags().Bool(cliflags.SyncOnStartFlag, true, cliflags.SyncOnStartFlagDescription)
-	_ = viper.BindPFlag(cliflags.SyncOnStartFlag, cmd.Flags().Lookup(cliflags.SyncOnStartFlag))
+	cmd.Flags().Bool(cliflags.SyncOnceFlag, false, cliflags.SyncOnceFlagDescription)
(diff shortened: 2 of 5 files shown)


cmd/dev_server/start_server.go
```diff
@@ -55,7 +58,7 @@ func startServer(client dev_server.Client) func(*cobra.Command, []string) error
 		if viper.IsSet(cliflags.ProjectFlag) && viper.IsSet(SourceEnvironmentFlag) {
 
 			initialSetting = model.InitialProjectSettings{
-				Enabled:    true,
+				Enabled:    viper.GetBool(cliflags.SyncOnStartFlag),
```
dmashuda, line 61: I think we are looking for behavior that is "only sync 1 time".

that way a customer could do:
```
ldcli dev-server start --project default --source staging --project default --source staging --sync-once true
```

That way the consumer does not need to vary their input for starting the server (without resyncing if the project/env exists)

mike-zorn: Yes. However, if we can make the cli flag be `--sync-once` instead of `--sync-once true`, I will be happier 😉

tvarney13: @mike-zorn I looked into it, it looks like the default behavior is what you want. However to achieve that [Cobra has some quirks](https://github.com/spf13/cobra/issues/1606).

tl;dr to set the flag to `false` you either have to not include the flag at all or use an `=` e.g. `--sync-once=false` which makes bool flags unique among other Cobra flags. [What I'm seeing in the code itself backs that up](https://github.com/launchdarkly/ldcli/blob/f2b610752f4afeb04e0ffcc92012aa1881f5ae93/vendor/github.com/spf13/pflag/bool.go#L56).

internal/dev_server/model/sync.go
```diff
@@ -29,16 +30,19 @@ func CreateOrSyncProject(ctx context.Context, settings InitialProjectSettings) e
 	var project Project
 	project, createError := CreateProject(ctx, settings.ProjectKey, settings.EnvKey, settings.Context)
 	if createError != nil {
-		if errors.Is(createError, ErrAlreadyExists) {
+		if !errors.Is(createError, ErrAlreadyExists) {
+			return createError
+		}
```
tvarney13, line 34: Moved the logic around to clean up the logic and hopefully make it easier to grok

internal/dev_server/model/sync.go
```diff
@@ -26,19 +27,21 @@ func CreateOrSyncProject(ctx context.Context, settings InitialProjectSettings) e
...
+
+		if settings.SyncOnce {
+			log.Printf("Project [%s] exists, but --sync-once flag is set, skipping refresh", settings.ProjectKey)
+		} else {
 			log.Printf("Project [%s] exists, refreshing data", settings.ProjectKey)
 			var updateErr error
-			project, updateErr = UpdateProject(ctx, settings.ProjectKey, settings.Context, &settings.EnvKey)
+			_, updateErr = UpdateProject(ctx, settings.ProjectKey, settings.Context, &settings.EnvKey)
```
tvarney13, line 41: Going to add a unit test for this

internal/dev_server/model/sync.go
```diff
@@ -48,6 +51,6 @@ func CreateOrSyncProject(ctx context.Context, settings InitialProjectSettings) e
 		}
 	}
 
-	log.Printf("Successfully synced Initial project [%s]", project.Key)
+	log.Printf("Successfully synced Initial project [%s]", settings.ProjectKey)
```
tvarney13, line 54: By using `settings.ProjectKey`, I was able to remove the dependency on the returned `project`. Any thoughts on a good replacement message since we're not syncing in some cases?

First thought: `Successfully initialized initial project [%s]` even though it's saying `initial` alot

tvarney13 (commented):
Going

internal/dev_server/model/sync.go
```diff
@@ -26,19 +27,21 @@ func CreateOrSyncProject(ctx context.Context, settings InitialProjectSettings) e
...
-	project, createError := CreateProject(ctx, settings.ProjectKey, settings.EnvKey, settings.Context)
+	_, createError := CreateProject(ctx, settings.ProjectKey, settings.EnvKey, settings.Context)
 	if createError != nil {
-		if errors.Is(createError, ErrAlreadyExists) {
+		if !errors.Is(createError, ErrAlreadyExists) {
+			return createError
+		}
+
+		if settings.SyncOnce {
+			log.Printf("Project [%s] exists, but --sync-once flag is set, skipping refresh", settings.ProjectKey)
```
mike-zorn, line 37: Should we return from here? Then you don't need to change the key stuff and there isn't the extra log line about syncing a project that we just said we skipped.

mike-zorn: I think that also changes the behavior to skip applying overrides from the CLI. I think that's actually what you want: don't apply overrides if we aren't syncing. Reason for that is that you already inited your dev server _with_ those overrides and you don't want to re-set them.

tvarney13: I opted to return early

cmd/config/testdata/help.golden
```diff
@@ -10,6 +10,7 @@ Supported settings:
 - `output`: Command response output format in either JSON or plain text
 - `port`: Port for the dev server to run on
 - `project`: Default project key
+- `sync-once`: Only sync new projects. Existing projects will not be resynced on startup
```
mike-zorn, line 13:
```suggestion
- `sync-once`: Only sync new projects. Existing projects will neither be resynced nor have overrides specified by CLI flags applied.
```

mike-zorn (approved):

Issue #490, opened 2025-02-07
Mobile Authorization Key & Project Value `api_key` Prefix (project not found)


matt-forster:
We are having issues connecting our mobile clients to the launch darkly development server. 

Specifically, the iOS build is prefixing the the value of the authorization header with `api_key`, so the development server is not recognizing the project. We are running into [this line](https://github.com/launchdarkly/ldcli/blob/b079f5399e9ae9b4669380eed5aaf24e6c6005e0/internal/dev_server/sdk/store_facade.go#L21).

Example:

1. development server is started with project named `example`.
2. The iOS client is sending requests in a way which the dev server is interpreting the project as `api_key example`.

We found the relevant swift code from the client:

```swift
    struct HeaderValue {
        static let apiKey = "api_key"
        static let applicationJson = "application/json"
        static let eventSchema3 = "3"
    }

    private let mobileKey: String
    private let additionalHeaders: [String: String]
    private let authKey: String
    private let userAgent: String
    private let wrapperHeaderVal: String?

    init(config: LDConfig, environmentReporter: EnvironmentReporting) {
        self.mobileKey = config.mobileKey
        self.additionalHeaders = config.additionalHeaders
        self.userAgent = "\(environmentReporter.systemName)/\(environmentReporter.sdkVersion)"
        self.authKey = "\(HeaderValue.apiKey) \(config.mobileKey)".
        ...
```

**To reproduce**

1. start the development server with an established project
2. route a mobile client using the URI config
3. make a flag request using the established project
4. observe the request failing in the dev server logs with the incorrect project

**Expected behaviour**

1. start the development server with an established project
2. route a mobile client using the URI config
3. make a flag request using the established project
4. the dev server recognizes the project correctly

**Logs**

```
::1 - - [07/Feb/2025:21:34:21 +0000] "GET /ui/ HTTP/1.1" 200 0 "" "Wget"
2025/02/07 21:34:23 project, api_key example, not found
2025/02/07 21:34:23 To add your project to the dev server, call `ldcli dev-server add-project --project api_key example --source {SOURCE_ENV_KEY}
172.18.0.1 - - [07/Feb/2025:21:34:23 +0000] "GET /meval/... HTTP/1.1" 404 33 "" "iOS/5.4.5"
```

**CLI version**

LDCLI 1.9.1

**Client Version**

We are currently in the process of upgrading to the latest, but we found this to be the case in
launchdarkly-react-native-client-sdk v8 and v9.

**OS/platform**

[...]

PR #495, opened 2025-02-10
fix: Handle api_key prefix in auth header

mike-zorn:
**Requirements**

**Related issues**

Closes https://github.com/launchdarkly/ldcli/issues/490

Supports the `api_key ` prefix on keys used for authentication.

### DIFF
--- a/internal/dev_server/db/sqlite.go
+++ b/internal/dev_server/db/sqlite.go
@@ -4,10 +4,11 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
-	_ "github.com/mattn/go-sqlite3"
-	"github.com/pkg/errors"
 	"io"
 	"os"
+
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/pkg/errors"
 
 	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
 	"github.com/launchdarkly/ldcli/internal/dev_server/db/backup"
@@ -380,6 +381,9 @@ func (s *Sqlite) RestoreBackup(ctx context.Context, stream io.Reader) (string, e
 
 func (s *Sqlite) CreateBackup(ctx context.Context) (io.ReadCloser, int64, error) {
 	backupPath, err := s.backupManager.MakeBackupFile(ctx)
+	if err != nil {
+		return nil, 0, errors.Wrapf(err, "unable to make backup file, %s", backupPath)
+	}
 	fi, err := os.Open(backupPath)
 	if err != nil {
 		return nil, 0, errors.Wrapf(err, "unable to open backup db at %s", backupPath)
--- a/internal/dev_server/sdk/project_key_middleware.go
+++ b/internal/dev_server/sdk/project_key_middleware.go
@@ -3,6 +3,7 @@ package sdk
 import (
 	"context"
 	"net/http"
+	"strings"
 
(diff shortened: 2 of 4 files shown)


internal/dev_server/sdk/project_key_middleware.go
```diff
@@ -38,6 +39,7 @@ func GetProjectKeyFromAuthorizationHeader(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
 		ctx := request.Context()
 		projectKey := request.Header.Get("Authorization")
+		projectKey = strings.TrimPrefix(projectKey, "api_key ") // some sdks set this as a prefix
```
mike-zorn, line 42: @keelerm84 Are there other scenarios we should be aware of for auth key formats?

keelerm84: I'm not sure. The `api_key` prefix isn't coming from any of the SDKs I wouldn't think. We don't add additional prefixing, outside of the standard key formats.

Maybe that's something more aligned with the integrations team, or whoever is in charge of the auto-generated API SDKs?

mike-zorn: lol glad we were both surprised by this behavior. It looks like iOS _is_ doing this [prefixing thing](https://github.com/launchdarkly/ios-client-sdk/blob/cb270ec1080ba7001b1892137f72151cf7e77be1/LaunchDarkly/LaunchDarkly/Networking/HTTPHeaders.swift#L34). Looks like same for [android](https://github.com/launchdarkly/android-client-sdk/blob/2a902c8f656a5ab4bb692d0b595c8e85f9796097/launchdarkly-android-client-sdk/src/test/java/com/launchdarkly/sdk/android/LDConfigTest.java#L102).

internal/dev_server/db/sqlite.go
```diff
@@ -380,6 +381,9 @@ func (s *Sqlite) RestoreBackup(ctx context.Context, stream io.Reader) (string, e
 
 func (s *Sqlite) CreateBackup(ctx context.Context) (io.ReadCloser, int64, error) {
 	backupPath, err := s.backupManager.MakeBackupFile(ctx)
+	if err != nil {
```
tvarney13, line 384: Oof nice catch - should we make a follow up item to add a linter rule to catch ineffectual assigns to `err`?

internal/dev_server/sdk/project_key_middleware.go
keelerm84: 🤯 Well that IS surprising. I wonder if that is still necessary.

I checked the relay, and there is a `fetchAuthToken` method that only checks for that `api_key` prefix, so maybe we can be more confident that prefix is the only one we have to worry about.

tvarney13 (approved):

internal/dev_server/sdk/project_key_middleware.go
mike-zorn: Ooh. Thank you for looking at the relay implementation. If that's all it supports, I think we're all good here.

internal/dev_server/db/sqlite.go
mike-zorn: The linter is running if you use pre-commit, but it doesn't look like pre-commit checks are being run in ci.

PR #520, opened 2025-04-14
feat: better error message when flag key is not found in the project

wichopy:
**Requirements**

**Related issues**

when using the CLI, its not clear that the error is the flag key does not match a key thats registered in the project. The error is just `"not found"`

**Additional context**

### DIFF
--- a/internal/dev_server/model/override.go
+++ b/internal/dev_server/model/override.go
@@ -2,7 +2,7 @@ package model
 
 import (
 	"context"
-
+	"fmt"
 	"github.com/launchdarkly/go-sdk-common/v3/ldvalue"
 )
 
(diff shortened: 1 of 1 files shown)


mike-zorn:
Thanks for the submission! That makes sense that we should distinguish between the project missing and the flag missing. We've used [error types to distinguish between 5xx & 4xx errors](https://github.com/launchdarkly/ldcli/blob/15cbf77e1291338b5a356d9694354014c5e39dd0/internal/dev_server/api/put_override_flag.go#L17-L16). As a result, the `fmt.Errorf` approach won't work. Instead we'll need to incorporate which kinds of things we didn't find in the custom error type and map that out in the API. Let me know if you wanna take that on. Otherwise, we can treat this as a feature request and implement it.

wichopy:
> Thanks for the submission! That makes sense that we should distinguish between the project missing and the flag missing. We've used [error types to distinguish between 5xx & 4xx errors](https://github.com/launchdarkly/ldcli/blob/15cbf77e1291338b5a356d9694354014c5e39dd0/internal/dev_server/api/put_override_flag.go#L17-L16). As a result, the `fmt.Errorf` approach won't work. Instead we'll need to incorporate which kinds of things we didn't find in the custom error type and map that out in the API. Let me know if you wanna take that on. Otherwise, we can treat this as a feature request and implement it.

Could i please just make a feature request then. When I was looking at the code I didn't see anything to reveal error codes with this `ErrNotFound` error so thoguht just adding the context to the error would be enough.

mike-zorn:
Oh. Yeah. We actually have a bug in here where we're returning 500s when it should be a 400. I'll close this one and fix up the error handling.

PR #555, opened 2025-06-05
🤞 fix release to dockerhub

mike-zorn:
This should fix two things

- Broken publishes to npm (remove spurious `always()` conditions)
- Dockerhub publish failures (log in to docker on the goreleaser container)


BehnH (approved):
🤞🏻

PR #561, opened 2025-06-09
fix(ci): [FUN-804] docker login command

mike-zorn:
This corrects the docker login command to not accept the password via stdin. The command needs to be done on the container, not on the host.

---
Related Jira issue: [FUN-804: Release new ldcli](https://launchdarkly.atlassian.net/browse/FUN-804)


BehnH (approved):
really hoping this is it 🤞🏻

PR #573, opened 2025-07-07
fix: correct CORS header name for exposing `date` in dev-server

Meligy:
fixes https://github.com/launchdarkly/react-client-sdk/issues/138

This fixes a provlem where the header `date` still fails in the browser even though it seems to be covered by `Access-Control-Expose-Headers` header:

image

The fix is simply removing an additional `:` that existed in the CORS header name.

This is hopefully a very straight forward fix, hence the short PR description.


dmashuda (approved):
Thank you for your contribution.

PR #614, opened 2025-09-10
Set CORS headers consistently for SDK & API cases

mike-zorn:
**Requirements**

- [ ] ~~I have added test coverage for new or changed functionality~~
  We don't have E2E tests for the dev server and this behavior can only be observed from the server being wired up.
- [x] I have validated my changes ~~against all supported platform versions~~

We observed that when `--cors-enabled` is set, some routes used by JS SDKs no longer allowed the correct set of headers under CORS. This solves that issue by altering how CORS middleware is applied. Instead of applying CORS middleware to the entire router, it is applied to subrouters so that the two CORS behaviors are accommodated. Additionally, this eliminates the custom CORS implementations for the one provided as part of gorilla.


internal/dev_server/api/api.yaml
```diff
@@ -7,9 +7,9 @@ info:
     those flags.
   version: 1.0.0
 servers:
-  - url: "http"
+  - url: "http://localhost:8765/dev"
```
mike-zorn, line 10: This is a somewhat unfortunate change, but this also seems better since the `/dev` prefix isn't repeated everywhere. I needed to remove the path from the routes so that I could use the `dev` prefix to apply the CORS middleware for the these routes.

PR #661, opened 2026-03-18
feat(REL-12754): Agent friendly error handling

nieblara:
Make every error message actionable: preserve the HTTP status code, and suggest what to do next. Today, many errors lose their status code entirely and users see "unknown error occurred."

Three phases: (1) stop losing status codes by always producing valid JSON errors, (2) add a suggestion registry mapping status codes to guidance, (3) update plaintext error formatting to include suggestions.

---

> [!NOTE]
> **Medium Risk**
> Touches core request/response formatting used across most CLI commands, so regressions could change user-visible output or automation expectations. Changes are mostly additive (status codes/suggestions, optional `--fields`, and improved plaintext formatting) with extensive test coverage.
> 
> **Overview**
> **Makes API errors agent-friendly.** HTTP error responses now always become valid JSON that includes `statusCode` (and, for common statuses like 401/403/404/409/429, an actionable `suggestion`), and plaintext error output prints the suggestion when present.
> 
> **Adds output shaping and better plaintext rendering.** Introduces global `--fields` to filter top-level keys in *JSON* output, and enhances plaintext output to render registered resources (e.g. `flags`, `members`, `projects`, `environments`, `segments`) as tables for lists and key-value blocks for singular results; commands now pass `ResourceName` into `CmdOutput` to enable this.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e027006885afbcb8ec16bd2ce04a2cf427cc596. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
Related Jira issue: [REL-12754: Better error messages with actionable suggestions](https://launchdarkly.atlassian.net/browse/REL-12754)

### DIFF
--- a/cmd/cliflags/flags.go
+++ b/cmd/cliflags/flags.go
@@ -13,6 +13,15 @@ func GetOutputKind(cmd *cobra.Command) string {
 	return viper.GetString(OutputFlag)
 }
 
+// GetFields returns the list of fields to include in JSON output, or nil if not specified.
+func GetFields(cmd *cobra.Command) []string {
+	fields, err := cmd.Root().PersistentFlags().GetStringSlice(FieldsFlag)
+	if err != nil {
+		return nil
+	}
+	return fields
+}
+
 const (
 	BaseURIDefault      = "https://app.launchdarkly.com"
 	DevStreamURIDefault = "https://stream.launchdarkly.com"
@@ -27,6 +36,7 @@ const (
 	DevStreamURIFlag = "dev-stream-uri"
 	EmailsFlag       = "emails"
 	EnvironmentFlag  = "environment"
+	FieldsFlag       = "fields"
 	FlagFlag         = "flag"
 	JSONFlag         = "json"
 	OutputFlag       = "output"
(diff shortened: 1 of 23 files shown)


JackDanger (commented):
Great improvement — actionable errors are one of the highest-leverage changes for both agents and humans. The suggestion registry is a smart pattern. Some things to consider:

**1. Error JSON shape change is a breaking contract for agent-prompts consumers**
The [launchdarkly-labs/agent-prompts](https://github.com/launchdarkly-labs/agent-prompts) repo (the official LD agent guide) teaches agents to check errors like this:
```bash
echo "$RESULT" | jq -e '.code == 401' > /dev/null
```
Before this PR, the `code` field could be a numeric-ish string or omitted. After this PR, it's always a snake_case string like `"unauthorized"`, and the numeric value moves to `statusCode`. This will break that exact pattern. The agent-prompts repo needs to be updated alongside this PR, or at minimum the PR description should call out the shape change explicitly.

**2. `baseURI` extraction from the request path is fragile**
```go
if parsed, err := url.Parse(path); err == nil && parsed.Scheme != "" {
    baseURI = parsed.Scheme + "://" + parsed.Host
}
```
If `path` is ever a relative path (e.g., `/api/v2/flags/my-proj`), `baseURI` will be empty, and the 401 suggestion will contain a broken link: `/settings/authorization` with no host. Worth adding a fallback to the configured `--base-uri` value, or at minimum a test that exercises the relative-path case.

**3. The suggestion for 404 could mislead**
> "Use `ldcli projects list` or `ldcli flags list` to see available resources."

If I'm working with segments, environments, metrics, or any of the ~375 auto-generated resource commands, this suggestion is wrong. Could the suggestion include the resource type from the command context? e.g., "Use `ldcli segments list`" when the failing command was `ldcli segments get`.

**4. Missing suggestion for 400 (Bad Request)**
This is arguably the most common error in practice — malformed JSON payloads, invalid field values, missing required fields. The API often returns helpful `message` text for 400s, but a suggestion like "Check the --data payload format. Use `ldcli <resource> create --help` for required fields." would help agents self-correct.

**5. Non-JSON error bodies get the raw HTML/text stuffed into `message`**
```go
"message": string(body),  // Could be "<html><body>Bad Gateway</body></html>"
```
This is tested and intentional, which is fine. But for agents, an HTML blob in the `message` field burns tokens. Consider truncating or sanitizing the message when the body isn't JSON (e.g., first 200 chars + `...`).

**6. Cursor Bugbot flagged 2 issues** — worth checking what those were before merging.

PR #660, opened 2026-03-18
feat!: agent-forward CLI — TTY-aware output, structured errors, --fields, table formatting [REL-12752]

nieblara:
Consolidates four agent-forward workstreams into a single PR. Originally stacked as #660 → #661 → #662 → #663, now merged into one branch for a clean merge to main.

## Breaking Change

**Default output is now JSON when stdout is not a TTY.** Scripts, CI, and agent environments that pipe `ldcli` output and relied on the plaintext default will now receive JSON.

Escape hatches (pick one):
```bash
ldcli flags list --project my-proj --output plaintext   # explicit flag
export LD_OUTPUT=plaintext                               # env var
export FORCE_TTY=1                                       # restore old default
```

Resolution order: `--json` > `--output` > `LD_OUTPUT` > config file > TTY-based default.

---

### 1. TTY-aware default output — [REL-12752](https://launchdarkly.atlassian.net/browse/REL-12752)

When stdout is not a terminal, `--output` defaults to `json` instead of `plaintext`. Follows the convention used by `gh`, `glab`, and `kubectl`. `FORCE_TTY` / `LD_FORCE_TTY` env vars restore the plaintext default (modeled on `NO_COLOR`).

Key files: `cmd/root.go`, `cmd/root_test.go`, `README.md`

### 2. Actionable error messages — [REL-12754](https://launchdarkly.atlassian.net/browse/REL-12754)

Errors now preserve the HTTP status code and include a `Suggestion:` line for common failures (401, 403, 404, 409, 429). Empty or non-JSON error bodies from the API are normalized into structured JSON with `code`, `message`, and `statusCode`.

Key files: `internal/errors/suggestions.go`, `internal/resources/client.go`, `internal/output/plaintext_fns.go`

### 3. `--fields` flag — [REL-12755](https://launchdarkly.atlassian.net/browse/REL-12755)

Filters JSON output to specified top-level fields. A single flag response can be 50KB+; `--fields key,name,kind` returns only those fields. Handles both singular and collection responses, preserves `totalCount` and `_links`, never filters error responses.

Key files: `cmd/cliflags/flags.go`, `internal/output/resource_output.go`

### 4. Table-formatted plaintext output — [REL-15756](https://launchdarkly.atlassian.net/browse/REL-15756)

Replaces `* name (key)` bullet output with aligned tables for list responses and key-value blocks for singular responses. Column definitions registered for flags, projects, environments, members, and segments. Unknown resources fall back to the existing bullet format.

[...]

PR #663, opened 2026-03-18
feat(REL-15756): updating agent friendly and improved rich text output

nieblara:
Replace the minimal plaintext output (* name (key) bullets) with table-formatted output using text/tabwriter. Define column sets per resource type starting with 5-6 high-value resources (flags, projects, environments, members, segments). Everything else falls back to the current bullet format. Singular resources show key-value pairs.

before:
Screenshot 2026-03-17 at 9 26 37 PM
Screenshot 2026-03-17 at 9 26 24 PM
Screenshot 2026-03-17 at 9 26 10 PM

after: 
Screenshot 2026-03-17 at 9 21 49 PM
Screenshot 2026-03-17 at 9 21 17 PM

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes the CLI’s default plaintext rendering across multiple resource commands, which can impact user workflows and tests that parse human-readable output (JSON output remains unchanged).
> 
> **Overview**
> Updates plaintext CLI output to be *resource-aware*: when `CmdOutput` is given a `ResourceName`, list responses render as aligned tables and singular responses render as key/value blocks (with fallback to the existing bullet and `name (key)` formats for unknown resources).
> 
> Adds `internal/output/table.go` with per-resource column registries (initially `flags`, `projects`, `environments`, `members`, `segments`) plus formatting helpers (yes/no booleans, tag truncation, timestamp formatting), and extends `CmdOutput` with optional `CmdOutputOpts` (including default `Fields`). Commands now pass `ResourceName` (e.g., `flags archive/toggle`, `members invite`, and generated `resources` operations), and tests are updated/expanded to assert the new plaintext output while keeping JSON behavior the same.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 887c6148bfd2494fd55aab0be1a19312429c6ab6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cmd/flags/archive.go
+++ b/cmd/flags/archive.go
@@ -51,7 +51,9 @@ func makeArchiveRequest(client resources.Client) func(*cobra.Command, []string) 
 			return output.NewCmdOutputError(err, cliflags.GetOutputKind(cmd))
 		}
 
-		output, err := output.CmdOutput("update", cliflags.GetOutputKind(cmd), res, cliflags.GetFields(cmd))
+		output, err := output.CmdOutput("update", cliflags.GetOutputKind(cmd), res, cliflags.GetFields(cmd), output.CmdOutputOpts{
+			ResourceName: "flags",
+		})
 		if err != nil {
 			return errors.NewError(err.Error())
 		}
--- a/cmd/flags/archive_test.go
+++ b/cmd/flags/archive_test.go
@@ -39,7 +39,11 @@ func TestArchive(t *testing.T) {
 
 		require.NoError(t, err)
 		assert.Equal(t, `[{"op": "replace", "path": "/archived", "value": true}]`, string(mockClient.Input))
-		assert.Equal(t, "Successfully updated test flag (test-flag)\n", string(output))
+		assert.Contains(t, string(output), "Successfully updated\n\nKey:")
+		assert.Contains(t, string(output), "test-flag")
+		assert.Contains(t, string(output), "Name:")
+		assert.Contains(t, string(output), "test flag")
+		assert.NotContains(t, string(output), "* ")
 	})
 
 	t.Run("succeeds with JSON output", func(t *testing.T) {
@@ -146,7 +150,9 @@ func TestArchive(t *testing.T) {
 		)
 
 		require.NoError(t, err)
-		assert.Equal(t, "Successfully updated test flag (test-flag)\n", string(output))
+		assert.Contains(t, string(output), "Successfully updated")
+		assert.Contains(t, string(output), "Key:")
+		assert.Contains(t, string(output), "test-flag")
 	})
 
 	t.Run("returns error with missing flags", func(t *testing.T) {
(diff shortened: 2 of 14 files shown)


JackDanger (commented):
The before/after screenshots look great — the table format is a significant UX improvement for humans. This is the kind of change that makes the CLI feel polished. A few things:

**1. PR title says REL-15756 but the branch is REL-12756**
Typo in the title — `feat(REL-15756)` should be `feat(REL-12756)`.

**2. This is a plaintext format breaking change**
Any script that parses the old bullet format (`* name (key)`) will break:
```bash
# This pattern stops working for flags, projects, environments, members, segments:
ldcli flags list --project my-proj -o plaintext | grep "^\* "
```
Research shows **no known consumers parse plaintext programmatically** (everyone uses `-o json`), so the practical risk is low. But it's still a format change for an existing output mode. Combined with #660's default-output change, this means scripts that neither set output format nor parse JSON get hit twice: different format AND different default.

**3. `CmdOutput` now takes 5 parameters — time to consolidate**
```go
CmdOutput(action, outputKind, input, fields, opts)
```
`fields` should live inside `CmdOutputOpts`. You already have the opts struct; having `fields` as a separate parameter alongside it is inconsistent. This would also clean up every call site that currently passes `nil, output.CmdOutputOpts{}`.

**4. Hardcoded column sets create a long tail of inconsistency**
You've defined table columns for 5 resource types (flags, projects, environments, members, segments). The CLI has ~375 auto-generated commands. Everything else falls back to the old bullet format. This means `ldcli flags list` shows a beautiful table but `ldcli segments list` shows bullets — wait, segments is covered. But `ldcli metrics list`, `ldcli contexts list`, `ldcli teams list` all fall back. Is there a plan to auto-generate column definitions from the OpenAPI spec, or will this be a gradual manual process?

**5. `formatTimestamp` drops the time component**
```go
// Parses RFC3339, outputs "2006-01-02"
```
For flags with recent changes, the date alone isn't very useful. "2026-03-17" tells me nothing about whether a flag was changed 2 minutes ago or 23 hours ago. Consider showing relative time for recent changes ("2h ago") and date for older ones, or at minimum including the time.

**6. Tag truncation magic number**
Tags are truncated at 3 items with `...`. This is fine UX, but the `3` is a magic number buried in the formatting function. A named constant would help readability.

**7. Singular resource key-value format is a nice touch**
The switch from `Successfully updated name (key)` to a proper key-value block with `Key:`, `Name:`, etc. is much better for both humans and agents reading plaintext. Good call.

---

**Cross-cutting concern for the whole stack:**

These 5 PRs together change: (1) default output format in non-TTY, (2) error JSON shape, (3) plaintext rendering format. The [agent-prompts repo](https://github.com/launchdarkly-labs/agent-prompts/blob/main/launchdarkly-cli.md) and [sandbox-data-generator scripts](https://github.com/launchdarkly/sandbox-data-generator) are the two known consumers to validate against. The agent-prompts guide specifically teaches `.code == 401` error checking which #661 changes.

Given ldcli uses semver and the v1→v2 bump was for an OpenAPI shape change, this stack probably warrants a v3 discussion — or at minimum a clear "Breaking Changes" section in the release notes.

ldhenry:
I'm in favor of doing a v3 for these changes as it gives us more flexibility in the design.

PR #706, opened 2026-05-11
feat: add command to check for updates

cspath1:
**Requirements**

**Related issues**

Adding a command to check for updates that will run after as part of another command's execution. Just a friendly little way to prompt the user to install a new version

---

> [!NOTE]
> **Overview**
> Adds an **automatic update notice** after normal `ldcli` runs: while the command executes, a background check compares the built-in version to the latest GitHub release and may print a short message on **stderr** when a newer version exists.
> 
> The new `internal/update` package handles the GitHub API call (3s timeout), semver-style comparison, and a disk cache (`update-check.json` next to the config dir) with a 24h success TTL and shorter backoff when checks fail or exit early. Users can disable checks via **`update-check-opt-out`** in `.ldcli-config.yml` / `ldcli config`; checks are also skipped when stderr is not a terminal. The root command waits up to **1 second** at shutdown (including before `os.Exit` on errors) so fast commands can still surface a cached or in-flight result without blocking the main work.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dda8d1910107ffe0714525706206c6c212916325. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
@@ -35,6 +36,7 @@ import (
 	"github.com/launchdarkly/ldcli/internal/members"
 	"github.com/launchdarkly/ldcli/internal/projects"
 	"github.com/launchdarkly/ldcli/internal/resources"
+	"github.com/launchdarkly/ldcli/internal/update"
 )
 
 type APIClients struct {
@@ -319,6 +321,19 @@ See each command's help for details on how to use the generated script.`, rootCm
 
 	rootCmd.cmd.SetUsageTemplate(getUsageTemplate())
 
+	// Start update check in the background so it runs in parallel with command execution.
+	type updateResult struct {
+		info *update.UpdateInfo
+	}
+	updateCh := make(chan updateResult, 1)
+	skipUpdateCheck := viper.GetBool(cliflags.UpdateCheckOptOut) ||
+		!term.IsTerminal(int(os.Stderr.Fd()))
+	if !skipUpdateCheck {
+		go func() {
+			updateCh <- updateResult{info: update.CheckForUpdate(version)}
+		}()
+	}
+
 	err = rootCmd.Execute()
 
 	var outcome string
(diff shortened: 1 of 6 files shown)


cmd/root.go
```diff
@@ -349,6 +364,16 @@ See each command's help for details on how to use the generated script.`, rootCm
...
 
 	analyticsClient.Wait()
+
+	if !skipUpdateCheck {
+		select {
+		case result := <-updateCh:
+			if result.info != nil && result.info.IsNewer {
+				fmt.Fprint(os.Stderr, update.NotificationMessage(result.info))
+			}
+		case <-time.After(500 * time.Millisecond):
```
LindseyB: Nit: Can we move this timeout into a constant

internal/update/update.go
```diff
@@ -0,0 +1,184 @@
...
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/launchdarkly/ldcli/internal/config"
+)
+
+const (
+	defaultGitHubReleasesURL = "https://api.github.com/repos/launchdarkly/ldcli/releases/latest"
+	cacheTTL                 = 24 * time.Hour
```
LindseyB, line 18:
```suggestion
	cacheTTL                 = 24 * time.Hour
	errorCacheTTL            = 1 * time.Hour
```

internal/update/update.go
```diff
@@ -0,0 +1,184 @@
...
+	client := &http.Client{Timeout: httpTimeout}
+	latest, err := fetchLatestVersion(client)
+	if err != nil {
+		return nil
+	}
+
+	writeCache(&cacheEntry{
+		LatestVersion: latest,
+		CheckedAt:     time.Now(),
+	})
```
LindseyB, line 169: I'm thinking we need to update this to make sure we don't hammer the GitHUb API if it's down or rate limiting us: 

```suggestion
  client := &http.Client{Timeout: httpTimeout}
  latest, err := fetchLatestVersion(client)
  if err != nil {
    writeCache(&cacheEntry{
        LatestVersion: currentVersion,
        CheckedAt:     time.Now().Add(errorCacheTTL - cacheTTL), // -23h, expires in 1h 
    })
    return nil
  }

  writeCache(&cacheEntry{
      LatestVersion: latest,
      CheckedAt:     time.Now(),
  })
```

LindseyB (commented):
Looks good, I just have some suggestions about how we can quickly handle rate limiting

internal/update/update.go
LindseyB: The suggested change box got funky but you should get the idea

nieblara:
@cursor Lindsey left some feedback here can you address them?

nieblara (approved):

PR #732, opened 2026-07-08
fix(dev-server): fetch flag pages concurrently instead of serially

davidbrackbill:
## Problem

The LD CLI dev server starts by fetching all flags in the project, which paginates the `GET /api/v2/flags/{projKey}` list endpoint at 100 flags/page serially, taking about ~50s in CI for Catamorphic.
-  [launchdarkly/gonfalon#66623](https://github.com/launchdarkly/gonfalon/pull/66623)

## Fix

- `GetPaginatedItems` now fetches page 0 to learn the total count and page size, then fetches the remaining pages **concurrently**, bounded to 6 in flight at once, and reassembles them in order.
- `Retry429s` is unchanged and still covers any individual page that gets rate-limited.

As measured locally, the full sync becomes ~50s → ~10s.

## Alternative considered and rejected: decouple `AvailableVariations` from server startup

I initially went further than pagination and additionally split `CreateProject`/`UpdateProject` so that only the fast SDK-streaming flag fetch (`AllFlagsState`) was synchronous, and the slow `GetAllFlags`-backed `AvailableVariations` fetch (used only by the local override-picker UI — never by the FD/streaming endpoints, which depend solely on `AllFlagsState`) ran in a background goroutine after the server started listening. That got healthcheck-ready time down to ~1-3s instead of ~10s.

## Risks

By adding parallel fetches, we risk increased load on this endpoint during startup, marginally. If several engineers/CI jobs run `dev-server start` around the same time against this shared account-wide project, the aggregate burst effect is worth a sanity check against the API's actual rate-limit policy before merging. 

## Testing

- `go test ./...` passes.
- `go test -race -count=3 ./internal/dev_server/...` passes.
- Verified live against the real staging `default` project (2,985 flags): full sync ~50s → ~10s, single run, no rate-limit errors observed.

---

> [!NOTE]
> **Medium Risk**
> Same request volume is compressed into shorter bursts against the LD API, which may increase 429s under shared load; mitigations are a concurrency cap and jittered retries, with no change to startup semantics or public API contracts.
> 
> **Overview**
> **Speeds up dev-server flag sync** by replacing serial `next`-link pagination with offset-based **concurrent page fetches** (up to 6 in flight) after the first response supplies `totalCount`.
> 
> `GetPaginatedItems` no longer takes an `href` or walks `_links.next`; `getFlags`/`GetAllFlags` callers keep the same public behavior. **`Retry429s`** adds up to 250ms **jitter** on top of `X-Ratelimit-Reset` waits (via new `MockableTime.Jitter`) so concurrent 429 retries don’t wake in lockstep. Tests cover ordering, concurrency, errors, and jittered retries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0d85c55ba00a2fecdbe22d485b4771f9c464d02. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/dev_server/adapters/internal/api_util.go
+++ b/internal/dev_server/adapters/internal/api_util.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"context"
 	"log"
+	"math/rand/v2"
 	"net/http"
 	"strconv"
 	"sync"
@@ -82,6 +83,12 @@ func GetPaginatedItems[T any, R interface {
 type MockableTime interface {
 	Sleep(duration time.Duration)
 	Now() time.Time
+	// Jitter returns a random duration in [0, max). Used to decorrelate
+	// retries after a 429: with concurrent page fetches, several requests
+	// can get rate-limited at once and would otherwise all compute the same
+	// X-Ratelimit-Reset and wake up to retry at the exact same instant,
+	// recreating the burst that got them limited in the first place.
+	Jitter(max time.Duration) time.Duration
 }
 
 type realTime struct{}
@@ -94,7 +101,19 @@ func (realTime) Now() time.Time {
 	return time.Now()
 }
 
+func (realTime) Jitter(max time.Duration) time.Duration {
+	if max <= 0 {
+		return 0
+	}
+	return rand.N(max)
+}
+
 var timeImpl MockableTime = realTime{}
+
+// maxRetryJitter bounds the random extra delay added on top of the API's
+// requested X-Ratelimit-Reset wait, so concurrent retries spread out instead
+// of waking up in lockstep.
+const maxRetryJitter = 250 * time.Millisecond
 
 func Retry429s[T any](requester func() (T, *http.Response, error)) (result T, err error) {
 	for {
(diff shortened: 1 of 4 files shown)


davidbrackbill:
Added jitter to the 429 retry backoff (\`b0d85c5\`) to address the burst/thundering-herd concern raised above: concurrent page fetches that get rate-limited around the same moment now retry at randomized offsets instead of all waking up at the exact same \`X-Ratelimit-Reset\` instant.

Testing path for this:
- Unit test simulates 8 concurrent goroutines all hitting a 429 with the *same* reset time, and asserts their resulting sleep durations differ (previously they'd have been identical).
- Existing single-retry test updated to assert the jitter is applied on top of the reset wait.
- No dedicated integration/load-test harness exists in this repo for exercising the real rate limiter under concurrency — verification beyond unit tests is manual (built the binary, ran it against the real staging project a few times, no 429s observed at concurrency=6). Flagging this as a gap rather than a guarantee.

internal/dev_server/adapters/internal/api_util.go
```diff
@@ -107,9 +126,9 @@ func Retry429s[T any](requester func() (T, *http.Response, error)) (result T, er
 				err = errors.Wrapf(err, `unable to retry rate limited request: X-RateLimit-Reset: "%s" was not parsable`, resetUnixMillisString)
 				return
 			}
-			sleep := resetUnixMillis - timeImpl.Now().UnixMilli()
-			log.Printf("Got 429 in API response. Retrying in %d milliseconds.", sleep)
-			timeImpl.Sleep(time.Duration(sleep) * time.Millisecond)
+			sleep := time.Duration(resetUnixMillis-timeImpl.Now().UnixMilli())*time.Millisecond + timeImpl.Jitter(maxRetryJitter)
```
davidbrackbill, line 129: Adds jitter in case of 429 retry to avoid thundering herds

dmashuda:
Would it be possible to source the required data for the initial sync from the streaming enpoint? (I am not sure if its possible, but if it is, it will be way faster)

https://github.com/launchdarkly/ldcli/blob/main/internal/dev_server/adapters/sdk.go#L41

davidbrackbill:
> Would it be possible to source the required data for the initial sync from the streaming endpoint? 

@dmashuda 
Ooh I like this idea, but that endpoint does not return variation names. Maybe we could lazy fetch those though? Not sure if it's a meaningful speed boost going from 10s to 1.5s

davidbrackbill:
closed, done in https://github.com/launchdarkly/ldcli/pull/733#issuecomment-4928771430

PR #756, opened 2026-07-24
feat(symbols): add Flutter (Dart AOT) symbol upload (--type flutter)

abelonogov-ld:
## Summary
- Add `ldcli symbols upload --type flutter` (aliases: `dart`). Parses Flutter ELF `app.<arch>.symbols` files with the Go stdlib `debug/elf` + `debug/dwarf`, extracts the Dart build id from `.note.gnu.build-id` and the DWARF debug info, and compiles them into the compact `.dartmap` symbol map (reusing the `dsymmap` codec from the Apple work, already on `main`).
- Uploads to both lanes: **Symbols Id** (keyed by build id) and **Version** (version + platform). Also wires Flutter into `symbols generate` for local `.dartmap` output.

## Test plan
- [ ] `go test ./internal/symbols/flutter/... ./cmd/symbols/...`
- [ ] `ldcli symbols upload --type flutter --path <split-debug-info dir>` and verify maps land in both lanes

Pairs with the backend Flutter enhancer (observability#1500) and Flutter SDK (observability-sdk#691) PRs.

Made with [Cursor](https://cursor.com)

PR #757, opened 2026-07-24
feat(symbols): upload Swift/ObjC sources with dSYMs (--include-sources)

abelonogov-ld:
Adds an opt-in `--include-sources` to `ldcli symbols upload --type apple-dsym`, which packs the source files an image's DWARF references into a `.srcbundle` and uploads it beside the `.dsymmap`. The backend then renders the code around each symbolicated frame instead of just `file:line`.

Stacks on #756 — merge that first and this diff reduces to the source-bundle work.

```
_sym/apple/id/<UUID>.dsymmap     # existing symbol map
_sym/apple/id/<UUID>.srcbundle   # new, same key
```

## What this adds

- `internal/symbols/srcbundle` — the Source Bundle format: header, an index sorted by path for binary search, a string table, and per-file gzip streams. Mirrors `.dsymmap`'s ethos and is byte-for-byte identical to the backend copy, with the same `Magic`/`Version` drift guard. Per-file compression means the backend inflates exactly one file to answer a frame, not the archive.
- Source collection — the existing DWARF walk now also records every source path it sees, from both line rows and inline call sites, resolving relative paths against the CU's `DW_AT_comp_dir`.
- `--include-sources` on both `symbols upload` and `symbols generate`.

## The key invariant

Files are keyed by the **exact** path string the sibling `.dsymmap` stores, because both artifacts are produced from one DWARF pass. A resolved frame's file name is therefore the lookup key, with no path normalization on either side.

## Safety and opt-in

Off by default, because it ships your source to LaunchDarkly; the flag help says so explicitly.

Packing skips anything not readable locally (SDK and system sources, or a dSYM archived on another machine), anything over 2 MiB, and any extension the UI can't render. If nothing local turns up it prints a note and uploads the map alone, so `--include-sources` never breaks an upload that would otherwise succeed.

## Testing

`./cmd/symbols/...` and `./internal/symbols/...` pass. New tests cover the format round trip, window clamping at file edges, files without a trailing newline, truncated/garbage input, keying by DWARF path rather than local path, skipping unreadable and oversize files, and that the real fixture dSYM (whose sources aren't on this machine) degrades to map-only instead of failing.

Verified end to end against a real iOS Release build.

Made with [Cursor](https://cursor.com)


Vadman97 (approved):
nice!

PR #765, opened 2026-07-30
feat(symbols): skip uploading symbols LaunchDarkly already has

abelonogov-ld:
## Summary

A rebuild that changes one module still re-sent every symbol file, so `ldcli symbols upload` took time proportional to the whole build rather than to what actually changed. That is most of the wall clock on a large R8 mapping or a fat dSYM.

`upload` now asks the backend for `skip_existing`. Any requested key the backend answers with an empty string is reported as already uploaded and skipped, and the summary becomes `Successfully uploaded all symbols (N already present)`. This covers all three upload paths: the generic React Native/Android loop, Apple dSYMs, and Flutter.

Dedup is on by default. It only applies where the key's id is derived from the bytes being uploaded, which the backend decides (launchdarkly/observability#1609): a Version Lane copy still overwrites, and so does a source bundle, since those are keyed by the id of the mapping or image they explain rather than by their own contents. `--no-skip-existing` forces a full re-upload, for repairing a truncated or corrupt stored object.

Because skipping is a change in what `upload` does by default, and the closing "Successfully uploaded all symbols" reads the same whether bytes were sent or not, a run that skipped anything now says so on **stderr** and names the flag. Someone re-uploading to repair a stored object would otherwise take a no-op run for a successful one. The notice is transitional and can come out after a release or two; the counts stay on stdout so scripts parsing them are unaffected.

**Backward compatibility:** a backend that predates the argument rejects the whole query at validation time, so the request is retried once without it. That keeps a freshly updated CLI working against an older or self-hosted deployment, at the cost of uploading bytes it might not have needed. Only the unknown-argument case is retried, so a credential or project error still reaches the user instead of being silently downgraded.

Depends on launchdarkly/observability#1609, which adds the argument. This should land after that deploys; until then the fallback path keeps `upload` working unchanged.

## How did you test this change?

New tests in `cmd/symbols/upload_dedup_test.go`, driving `getSymbolUploadUrls` against an `httptest` backend that records each request:

- the dedup document and `skip_existing: true` are sent when enabled, and neither is sent with `--no-skip-existing`
- an `Unknown argument "skip_existing"` response triggers exactly one retry, and the retry drops the argument
- any other GraphQL error is returned without a retry
- `alreadyUploaded` treats only the empty string as a skip
- the summary prints the skipped count on stdout and the notice on stderr, and stays silent on stderr when nothing was skipped

`go build ./...`, `go vet ./cmd/symbols/...`, and `go test ./cmd/symbols/...` pass.

---

> [!NOTE]
> **Medium Risk**
> Changes default upload behavior and presigned-URL handling across all symbol types; mis-dedup could leave stale symbols, mitigated by opt-out flag and backend digest checks for non-content-keyed bundles.
> 
> **Overview**
> **`ldcli symbols upload` skips re-sending symbols the backend already stores**, so repeat uploads after unchanged builds avoid large PUTs. Dedup is **on by default** across the generic React Native/Android path, Apple dSYM, and Flutter flows. The CLI asks `get_symbol_upload_urls_ld` with **`skip_existing`**; an **empty presigned URL** means skip that key. **`--no-skip-existing`** forces a full re-upload for repair scenarios.
> 
> **Source bundles** (Apple/Android `--include-sources`) are keyed off the mapping/image id, not bundle bytes, so skips use parallel **MD5 digests** (S3 ETag form) so the backend can match stored content. **Flutter Version-lane** copies still always get URLs so `--app-version` overwrites keep working; only Id-lane keys can skip.
> 
> **Older backends** that reject the new GraphQL args get **one retry** without dedup (only on unknown-argument errors). Completion messaging reports **`(N already present)`** on stdout and a **stderr** note naming `--no-skip-existing` when anything was skipped.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a7418b388163cfbbcea80077415aa1806b7c40b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/symbols/apple_upload.go
+++ b/cmd/symbols/apple_upload.go
@@ -22,6 +22,10 @@ const appleSymbolsIDPrefix = "_sym/apple/id"
 // in lookup (maps are keyed by UUID); the backend appends the same extension.
 const appleSymbolExt = ".dsymmap"
 
+// kindSources marks an appleSymbolMap that carries packed sources rather than a
+// symbol map.
+const kindSources = "sources"
+
 // appleSymbolMap is one compiled artifact ready to upload: an architecture's
 // .dsymmap symbol map, or (with --include-sources) its .srcbundle sources.
 type appleSymbolMap struct {
@@ -48,9 +52,9 @@ func (m appleSymbolMap) label() string {
 // context around native frames.
 //
 // With skipExisting, an unchanged binary keeps its UUID, so re-running this sends
-// nothing. Source bundles are exempt (the backend decides): they are keyed by their
-// image's UUID rather than their own contents, so sources that were unreadable on the
-// machine that uploaded first still need to overwrite.
+// nothing. Source bundles borrow that UUID rather than being keyed by their own
+// contents — sources unreadable on the machine that uploaded first must still be able
+// to overwrite — so those are skipped only when their digest matches what is stored.
 func uploadAppleDSYMs(apiKey, projectID, path, backendURL string, includeSources, skipExisting bool) error {
 	images, err := findDSYMImages(path)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


Vadman97 (approved):
nice optimization!

PR #768, opened 2026-07-31
feat(symbols): build the R8 mapping index in the CLI

abelonogov-ld:
## Summary

`symbols upload --type android` sent R8's `mapping.txt` as it was — 61.3 MB for the e2e app — and the backend parsed it into a random-access index on the first crash of every build that ever crashed. The mapping is here, on the machine that just produced it, so the index is built here and the text is never stored: `upload` and `generate` now write `mapping.v1.index` to both the Symbols Id Lane and the Version Lane. 61.3 MB of text becomes a 5.2 MB index.

**Streamed, not parsed.** `r8index.EncodeFrom` reads the mapping line by line and encodes each class as it closes, and `srcbundle.Builder.Add` now gzips an entry on add, so a per-class bundle costs its compressed size rather than its raw one. Peak live heap for that mapping is 100.3 MB, down from 407.5 MB parsing it whole — a build machine no longer has to find hundreds of megabytes to produce a few.

**A mapping that yields no index is an error.** Symbolication reads only the index, so storing anything else would leave a build looking like it has symbols while every crash arrives obfuscated.

**`internal/symbols/r8index` is a verbatim copy of the backend's package**, the way `dsymmap` and `srcbundle` already are. An index is written by one repo and read by another, so the copies are only interchangeable while they agree byte for byte: the golden fixture — a readable mapping and the exact index bytes it encodes to — is identical on both sides, so drift fails a test here rather than surfacing as an unreadable index or a silently wrong frame. The format version is in the artifact's file name, so an index written by an older CLI is a key the reader never asks for rather than bytes it has to make sense of.

Android upload logic moves to `cmd/symbols/android_upload.go`, and the generic upload path no longer carries Android special cases.

## Ordering

Stacked on #767.

The backend read side is https://github.com/launchdarkly/observability/pull/1612, and it must be deployed **after** this ships: it stops reading `mapping.txt`, and this is what starts writing the index it reads instead. Between the two, an Android build uploads an index nothing reads yet. Any lane holding only `mapping.txt` needs re-uploading once the backend flips.

## Test plan

- [x] `go test ./cmd/symbols/... ./internal/symbols/...`
- [x] Golden index bytes identical to the backend's copy of the fixture
- [x] The streaming and whole-mapping encoders produce the same bytes
- [x] `R8_MAPPING`/`R8_INDEX` cross-check: an index this CLI built answers like the 61.3 MB mapping it came from
- [x] Local end-to-end: e2e release build uploaded, crash symbolicated from the index with no `mapping.txt` in the lane

Made with [Cursor](https://cursor.com)


Vadman97 (approved):
fancy

PR #787, opened 2026-09-01
fix: redact access token in config command output

### LINKED ISSUE #786: config --list prints the stored access token in full
**Is your feature request related to a problem? Please describe.**

`ldcli config --list` renders the stored `access-token` value verbatim, in every output mode:

```console
$ ldcli config --list
{"access-token":"api-<the real token>","environment":"production","project":"my-project"}
```

Reproduced against 3.9.0 built from `main`.

Two things make this sharper than it first looks:

1. JSON is the default whenever stdout is not a terminal (`cmd/root.go:227-230`), so the piped and CI-log path is the one that leaks by default.
2. `config --list` is the natural way to confirm that `--set` landed, so it is a command people are led to run right after storing the token.

The result is the token in terminal scrollback, shell-session captures, screen shares, and CI logs. I hit this in practice: the value ended up in a transcript and the token had to be treated as compromised and rotated.

A second, narrower instance of the same problem is in the `--set`/`--unset` validation error, which echoes back the argument it rejected:

```console
$ ldcli config --set api-<the real token> access-token
{"message":"api-<the real token> is not a valid configuration option"}
```

Transposing the key and value is an easy mistake, and it puts the secret in the error message.

To be explicit about scope: **this is not a vulnerability report and I am not filing it as one.** No privilege boundary is crossed and no third party gains anything — the CLI is printing the operator's own local secret, on request, to their own terminal. It is a secret-hygiene / defense-in-depth issue. I read SECURITY.md and did not use Bugcrowd for that reason; happy to be redirected if you disagree.

**Describe the solution you'd like**

Redact the value in output, unconditionally, and leave the stored value untouched:

```console
$ ldcli config --list
{"access-token":"[REDACTED]","environment":"production","project":"my-project"}
```

`config --list` answers "is a token configured, and what are my other defaults" — and a redacted marker answers that just as well as the value does. Nothing in the CLI needs the value echoed back to it.

- **Redact by default with a `--show-secrets` opt-in.** Preserves the capability, but it weakens the guarantee to "safe unless someone passes a flag", and a flag that exists to print a secret invites being pasted into scripts and automation. I would rather not add it unless there is a use case that needs it.
- **Partial masking (`api-••••••••cd12`).** Helps a human tell which of several tokens is stored. It still puts characters of a live secret into scrollback, and I do not think the disambiguation is worth that — `ldcli whoami` already names the token without revealing it.
- **Leaving it alone because the config file is readable anyway.** Reading `$XDG_CONFIG_HOME/ldcli/config.yml` is a deliberate act against a file with a known path. Printing to stdout is passive and ends up in logs and transcripts that outlive the terminal. Those are different exposure profiles.

The config file also means **no capability is lost** by redacting: anyone who genuinely needs the value can still `yq '.["access-token"]' "$XDG_CONFIG_HOME/ldcli/config.yml"`.

**Additional context**

I have a PR ready that does both of the above, with tests, and I will link it here.

rchaves-nexar:
**Requirements**

**Related issues**

Fixes #786.

The stored access token was rendered verbatim by `config --list` and by the `--set`/`--unset` validation error. Two commits, one per path.

**1. `config --list`.** Adds `Config.Redacted()`, which returns a copy with sensitive values replaced by `[REDACTED]`, and marshals that for output instead of the raw config. The plaintext renderer, `--output json`, and the JSON default used when stdout is not a terminal all derive from the same `json.Marshal` in `cmd/config/config.go`, so the single substitution covers all three.

Two details worth flagging for review:

- The receiver is a value, so the copy is free and nothing that is written back to disk is affected. Verified end to end: the config file is byte-identical after `--list`.
- An *unset* token is left empty rather than set to `[REDACTED]`, so `omitempty` still elides the key. Reporting an absent token as "present but hidden" would be a worse answer than omitting it.

I chose an explicit method over reflecting on a new struct tag, or over unmarshalling to a `map` and substituting there. The map route re-sorts the keys and changes output shape for everyone; the method preserves field order and matches the hand-written `switch` already used in `Update`.

**2. Validation errors.** `--set` and `--unset` reject an unknown key by echoing it back. The argument in that position is not always a key — transposing `--set <key> <value>` puts the value there, and for `access-token` that value is a secret:

```console
$ ldcli config --set api-<token> access-token
{"message":"api-<token> is not a valid configuration option"}
```

Rather than trying to detect secrets — a heuristic on token prefixes will rot the moment a token format changes — this checks the *shape* of what is about to be echoed. Configuration keys are short lowercase-hyphenated words; anything else is not a key, so there is no reason to reflect it back. The typo case that the echo exists to serve is unaffected:

```console
$ ldcli config --set projct foo
{"message":"projct is not a valid configuration option"}     # unchanged

$ ldcli config --set api-<token> access-token
{"message":"[REDACTED] is not a valid configuration option"}  # fixed
```

- **A `--show-secrets` flag.** Deliberately omitted. It weakens the guarantee to "safe unless someone passes a flag", and a flag whose purpose is to print a secret tends to end up in scripts. No capability is lost without it: the config file is plain YAML at a documented path, so `yq '.["access-token"]' "$XDG_CONFIG_HOME/ldcli/config.yml"` still works for anyone who genuinely needs the value. Happy to add the flag if you would rather not take the behavior change unconditionally.
- **Partial masking (`api-••••••••cd12`).** Lets a human identify which token is stored, but still puts live secret characters into scrollback, and `ldcli whoami` already names the token without revealing it.
- **A registry-driven sensitive-key set in `cmd/cliflags`.** Attractive since `AllFlagsHelp()` is already the single authority for valid config keys, but promoting it from `map[string]string` to a struct changes a shape used by the public `config --help` listing, for no benefit at one sensitive key. Easy to move to later if a second one appears.

**Additional context**

Behavior change, so calling it out plainly: any caller parsing the token out of `config --list` output will now read `[REDACTED]`. I believe that is the point of the change rather than a regression, but it is your call whether it needs more than a patch release.

`cmd/config/testdata/help.golden` is unchanged — no flag is added and no description is edited.

Tests added in `internal/config/config_test.go`:

- `Redacted()` — token replaced when set; left empty when unset (asserted through `json.Marshal`, so `omitempty` is actually exercised); non-sensitive fields untouched; receiver not mutated.
- `TestRedactedOutput` — marshals the redacted config through `output.CmdOutputSingular` for both `json` and `plaintext`, asserting the token appears in neither. This is what pins sites 2 and 3 rather than just the struct.
- `TestErrorDoesNotEchoNonKeyArguments` — transposed `--set` and non-key `--unset` redact; a key-shaped typo still echoes verbatim.

I skipped an end-to-end `config --list` test on purpose: it needs `viper.SetConfigFile` on global state that `cmd/root.go` owns, and a flaky test seemed worse than a missing one. I did verify the built binary by hand in all three output modes.

Verified locally: `make build`, `go test ./...`, `go vet`, and `golangci-lint` v1.63.4 all clean; `gofmt -l` reports nothing in the touched files. (`cmd/root.go` and `cmd/dev_server/projects.go` are flagged by gofmt on `main` already — I left them alone rather than mix unrelated formatting into this diff.)

---

> [!NOTE]
> **Overview**
> Prevents the LaunchDarkly access token from leaking in **`ldcli config`** output and validation errors (fixes #786).
> 
> **`config --list`** now serializes `Config.Redacted()` so a stored token appears as **`[REDACTED]`** in JSON and plaintext; unset tokens stay omitted via `omitempty`. On-disk config is unchanged.
> 
> **`--set` / `--unset`** invalid-key errors use **`safeKeyForError`**: only short, hyphenated key-shaped strings are echoed (typos like `projct` still show); anything else—including a token accidentally passed where a key is expected—is replaced with **`[REDACTED]`** in the message.
> 
> Tests cover redaction, output rendering, and error echo behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6982f9e32e1511c821837df7c0a469e0c73fd4c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rchaves-nexar:
Hey @ffantl-ld , thanks for taking a look. Any thoughts on the approach? Is there a better way?

ffantl-ld:
Hi @rchaves-nexar - I'm actually happy with the approach you took. I've been thinking about the [REDACTED] versus anything else, but I think this might just be the right approach for now. We might change it to a separate approach since we sometimes do want these to be readable.

ffantl-ld (approved):

rchaves-nexar:
Thanks @ffantl-ld . I would usually hit the "Merge" button, but that's not an option here. Any further input or action required from my side?

Thanks again for taking the time :pray:

ffantl-ld:
Thanks for contributing!

PR #794, opened 2026-09-11
feat(sync): compile local prompt variations

ctawiah:
## Summary

Compiles prompt variation files across projects in a `.launchdarkly` workspace into deterministic sync input. This layer builds on the variation parser in the preceding PR.

- Walks each project configuration directory and ignores unsupported files.
- Sorts variations by project and lookup key for stable request construction.
- Includes the failing file path in parse errors and reports a missing `.launchdarkly` directory clearly.

## Verification

- `go test ./internal/sync/... -count=1`
- Focused lint for the changed sync packages

### DIFF
--- a/internal/sync/local/compile.go
+++ b/internal/sync/local/compile.go
@@ -1,0 +1,112 @@
+package local
+
+import (
+	"cmp"
+	"errors"
+	"io/fs"
+	"path"
+	"slices"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


ctawiah:
Consolidated into #788 so prompt variation parsing and compilation can be reviewed as one coherent local-loading change.

PR #796, opened 2026-09-11
feat(sync): persist ldcli installation identity

ctawiah:
## Summary

Persists a stable installation ID in the existing ldcli user configuration for non-Git source identity. This is a focused configuration layer above Git source detection.

- Creates a random installation ID only when one does not already exist.
- Writes the config atomically with restrictive permissions and preserves existing and unknown settings.
- Keeps the internal ID out of `ldcli config --list` output.
- Adds no metadata file to the customer workspace.

## Verification

- `go test ./internal/config -count=1`
- Focused lint for `./internal/config`


ctawiah:
Consolidated with the other automatic source identity work in #789.

PR #797, opened 2026-09-11
feat(sync): derive automatic local source identity

ctawiah:
## Summary

Adds automatic source selection for workspaces that do not have a usable Git origin. This layer combines Git discovery with the persisted installation identity.

- Prefers normalized Git source identity when it is available.
- Falls back to `sha256(installation ID + canonical workspace path)` without exposing the local path.
- Uses the nearest `.launchdarkly` parent as the workspace root, or the current directory before bootstrap.
- Normalizes symlinks, adds no source metadata file, and exposes no source-selection flag.

Moving a non-Git workspace intentionally creates a new local source identity.

## Verification

- `go test ./internal/sync/... -count=1`
- Focused lint for `./internal/sync/source`

### DIFF
--- a/internal/sync/source/resolver.go
+++ b/internal/sync/source/resolver.go
@@ -1,0 +1,112 @@
+package source
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"path/filepath"
+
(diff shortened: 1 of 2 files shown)


ctawiah:
Consolidated with Git detection and installation identity in #789.

